PR #35, opened 2023-01-31
[NO-CSL] Export ESM and CJS

esezen:
We might need to add an extra script to add individual package.json files to the folders under lib but let's try this first. Script should be pretty simple if we need it


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.1.5",
   "description": "Constructor.io Autocomplete UI library for web applications",
   "main": "lib/src/index.js",
-  "module": "lib/src/index.js",
```
esezen, line 6: I don't think we need `module` since we are only supporting ESM at the moment

esezen: Stale

sblaurock (approved):

PR #38, opened 2023-02-01
update verify node script and run npm audit fix with node lts

mocca102:


esezen:
Can we also update the GH workflows to make sure they use the same Node version we do?

package.json
```diff
@@ -23,7 +23,8 @@
     "test-storybook:ci": "start-server-and-test storybook:ci http://localhost:6006 test-storybook",
     "build-storybook": "storybook build -o 'docs'",
     "serve-built-storybook": "npx http-server docs",
-    "version": "npm run build-storybook && git add -u ./docs && git add ./docs/* && npm run compile",
+    "verify-node-version": "chmod +x ./scripts/verify-node-version.sh && ./scripts/verify-node-version.sh",
+    "version": "npm run verify-node-version && npm run build-storybook && git add ./docs/* && npm run compile",
```
esezen, line 27: Is there a reason why we changed the git add command here?

mocca102: I don't remember so it's mostly a mistake I'm putting it back

PR #37, opened 2023-02-01
Psl 1609 os UI ac update default style docs

deomsj:
- Update Storybook and readme to illustrate new workflow for opting into default styles
 - update import statements in docs to import from new library name:
   - instead of: from 'cio-autocomplete-ts'
   - use:  from '@constructor-io/constructorio-client-javascript'
 - and just for funsies, replace sb with cio icons

PR #39, opened 2023-02-02
[CSL-2053] Update Storybook to fix manager.js issue

esezen:
Reintroduces the changes we made here: https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/37
We had to revert those changes due to a Storybook bug


.gitignore
```diff
@@ -39,7 +39,7 @@ bower_components
 build/Release
 
 # Dependency directories
-/node_modules/
+node_modules/
```
deomsj, line 42: we don't need the leading `/` anymore?  no more node_modules buried in the docs folder?

deomsj (approved):
nice detective work on this, @esezen! 

Looks great from local testing, excited to put this to the test and see how it plays with github pages :)

PR #40, opened 2023-02-15
improve input enter key handling

deomsj:
Currently, when the autosuggest menu is open and a keyboard user has not key'd down or up to highlight a suggestion, then they press enter straight from the input field, the menu closes but the term is not actually submitted. This is not what users would expect. Let's fix that!

Corrected Behavior with this code change:
- when the autosuggest menu is open (and the user has not key'd down or up to highlight a suggestion), then they press enter from the input field:
  - close the menu
  - submit the search term
  - send search submit tracking event
  - when search term is empty, do not fire off a search submit or tracking events


esezen (approved):
LGTM! Thanks for catching this and fixing it 🚀

PR #50, opened 2023-03-02
Psl 1619 dont show sections that have 0 results

deomsj:
- Currently we will still show a section header even if there are no results to display.
 - In order to prevent confusion, we should not show a section with 0 results

Before this PR:
Screenshot 2023-02-02 at 6 32 36 PM

With changes in this PR:
Screenshot 2023-03-02 at 10 23 47 AM
Screenshot 2023-03-02 at 10 23 39 AM
Screenshot 2023-03-02 at 10 13 24 AM
Screenshot 2023-03-02 at 10 13 14 AM


esezen (approved):
LGTM! Thanks for working on this

PR #51, opened 2023-03-03
[CSL-2130] Export Types from Root Path

Mudaafi:
Currently, package consumers must import types via:
`@constructor-io/constructorio-ui-autocomplete/lib/types/types`. 

What we really want is the ability to do:
`import type { SectionConfiguration } from '@constructor-io/constructorio-ui-autocomplete'`.


src/index.ts
```diff
@@ -4,3 +4,4 @@ export { default as SearchInput } from './components/Autocomplete/SearchInput/Se
 export { default as SectionItem } from './components/Autocomplete/SectionItem/SectionItem';
 export { default as SectionItemsList } from './components/Autocomplete/SectionItemsList/SectionItemsList';
 export { default as useCioAutocomplete } from './hooks/useCioAutocomplete';
+export * from './types';
```
mocca102, line 7: Brilliant!

mocca102 (approved):

PR #52, opened 2023-03-03
Psl 1692 simplify interface for getItemProps

deomsj:
We identified `sectionIdentifier` should not default to Products and it should be a required parameter (getItemProps)
 - Remove the default value for sectionIdentifier
     - ✅ done in previous commit, then removed as it is no longer needed!!!
 - console.error  if this is not passed (or if there is a better way to warn the developer)
    - ✅ done in previous commit, then removed as it is no longer needed!!!
 - Update appropriate documentation
    - ✅ done
 - Look for other variables that have a similar issue and do the same
    - ✅ none of the other propGetter callback functions accept any arguments
 - remove index parameter
     - ✅ done
 - Update documentation accordingly
     - ✅ done
 - Update stories accordingly
     - ✅ done


mocca102 (approved):
Thank you for the changes 🚀

PR #61, opened 2023-05-05
Add dependabot.yml

jjl014:
### Updates:
* Configure dependabot to check for security PRs daily and add reviewers


esezen (approved):
LGTM!

PR #67, opened 2023-06-09
[CSL-2463] Resolve "Code Scanning" warnings

esezen:
Resolves security warnings caused by us and also simplifies description generation


src/utils.ts
```diff
@@ -63,26 +63,30 @@ ${templateCode}
   };
 };
 
-export const defaultOnSubmitCode = `"onSubmit": (submitEvent) => console.dir(submitEvent)`;
-
-export const defaultArgumentsCode = (apiKey: string) => `"apiKey": "${apiKey}",
```
mocca102, line 68: This is used [here](https://github.com/Constructor-io/constructorio-ui-autocomplete/blob/main/src/stories/Autocomplete/Hook/UserEvents.stories.tsx#L4) and [here](https://github.com/Constructor-io/constructorio-ui-autocomplete/blob/main/src/stories/Autocomplete/Component/UserEvents.stories.tsx#L4)  

I can remove them and update the args if you'd like to

src/utils.ts
```diff
@@ -63,26 +63,30 @@ ${templateCode}
...
-  ${defaultArgumentsCode(apiKey)}`
+  // Stringify non-function values normally. Add a template block for functions to be replaced later
+  let res = JSON.stringify(
+    obj,
+    (key, value) => (value instanceof Function ? `${key}_CODE` : value),
+    '  '
   );
+
+  // Replace template blocks with function strings
+  Array.from(res.matchAll(/"(\w*)_CODE"/g)).forEach((match) => {
```
mocca102, line 79: Big brains

mocca102 (changes requested):
It's looking good except for the places where `defaultArgumentsCode` that got removed was used

src/utils.ts
esezen: Good catch! Fixed here [13dab01](https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/67/commits/13dab0165f37c53be0c4872598acb8de960a4343)

mocca102 (approved):
LGTM!

PR #68, opened 2023-06-14
[CSL-2454] Fix cursor jump when editing query

stanlp1:
Fixed according to this github post. 

https://github.com/downshift-js/downshift/issues/1108#issuecomment-674180157


mocca102 (approved):
LGTM!

PR #72, opened 2023-06-27
[CSL-2527] Add support to passing in autocomplete parameters

stanlp1:
It made sense to me to just allow users to pass in an object under autocomplete parameters.


src/types.ts
```diff
@@ -2,12 +2,14 @@
...
 import { ReactNode } from 'react';
 import ConstructorIOClient from '@constructor-io/constructorio-client-javascript';
+import { IAutocompleteParameters } from '@constructor-io/constructorio-client-javascript/lib/types';
 
 export type CioClientConfig = { apiKey?: string; cioJsClient?: ConstructorIOClient };
 
 export type AdvancedParameters = {
   numTermsWithGroupSuggestions?: number;
   numGroupsSuggestedPerTerm?: number;
+  autocompleteParameters?: Omit<IAutocompleteParameters, 'resultsPerSection'>;
```
deomsj, line 12: well played with this Omit generic :)

src/stories/Autocomplete/Component/AdvancedParameters.stories.tsx
```diff
@@ -49,3 +50,21 @@ addComponentStoryDescription(
...
 );
+
+export const TermsWithFilters = ComponentTemplate.bind({});
+TermsWithFilters.args = {
+  apiKey,
+  onSubmit,
+  advancedParameters: {
+    autocompleteParameters: {
+      filters: {
+        'Products][color': 'Black',
```
deomsj, line 61: whoa, this key name is trippy `Products][color`

deomsj:
Yes! Excited to see this progress, @stanlp1. Did you notice this failing test? You down to investigate, or are you stuck on this and want help?

failing test

src/constants.ts
```diff
@@ -202,3 +202,7 @@ export const advancedParametersDefaultDescription = `Passing an \`advancedParame
...
 - type "bab" in the example below. Notice how the user is presented with a search term of "baby" as well as "baby in Bodysuits" and "baby in Baby Organic Cotton"
 - navigate to the "Terms With Group Suggestions" story (using the navigation menu to the left), then use the Controls to adjust the values of \`numTermsWithGroupSuggestions\` to \`3\` and \`numGroupsSuggestedPerTerm\` to \`1\`. Next, type "dan" in the example autocomplete input field. Notice how the user is presented with three different search terms that have a maximum of one "in {group}" suggestion each`;
+
+export const termsWithFiltersDescription = `Pass a autocompleteParameters filter object under advancedParameters. This will allow you to filter results down. To see this in action:
```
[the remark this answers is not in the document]
esezen: Would it be possible to make this description more generic? Something like "You can pass any parameters supported in client-js with this prop. For example filters......."

src/stories/Autocomplete/Component/AdvancedParameters.stories.tsx
esezen, line 61: Can we use a different example for now? ][ action might confuse users.
We can 
* use group_id for filters because Search Suggestions have group IDs
* contact DS to add facets to Search Suggestions
* showcase a different parameter than filter

src/stories/Autocomplete/Component/AdvancedParameters.stories.tsx
```diff
@@ -49,3 +50,21 @@ addComponentStoryDescription(
   `const args = ${stringifyWithDefaults(TermsWithGroupSuggestions.args)}`,
   termsWithGroupSuggestionsDescription
 );
+
+export const TermsWithFilters = ComponentTemplate.bind({});
```
esezen, line 54: Can we rename this? It makes me think Search Suggestions will be filtered but it doesn't do that. Not sure what a better name could be though 🤔

esezen (commented):
Left some small comments but this is looking great! We can merge after Jesse gives another review.

Can you also merge in main? It looks like another PR touched the same test

src/stories/Autocomplete/Component/AdvancedParameters.stories.tsx
deomsj: would love it if we could provide an example that leverages group_ids as well as an example that leverages a different facet with standard/non-distracting key names

stanlp1: Good points, opted to use a group_id filter instead.

I don't think any of the other autocomplete parameters currently make too much sense to show as an example so I stayed with filters.

src/constants.ts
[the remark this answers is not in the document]
stanlp1: 👍

src/stories/Autocomplete/Hook/AdvancedParameters.stories.tsx
```diff
@@ -49,3 +50,21 @@ addHookStoryCode(
...
+export const FilteredSuggestions = HooksTemplate.bind({});
+FilteredSuggestions.args = {
+  apiKey,
+  onSubmit,
+  advancedParameters: {
+    autocompleteParameters: {
+      filters: {
+        group_id: '1130',
+      },
+    },
```
deomsj, line 63: as a consumer of this lib, I'd find this to be confusing:

instead of:
```js
advancedParameters: {
    autocompleteParameters: {
      filters: {
        group_id: '1130',
      },
    },
```

I'd prefer:
```js
advancedParameters: {
  filters: {
    group_id: '1130',
  },
},
```

src/hooks/useDebouncedFetchSections.ts
```diff
@@ -83,7 +83,7 @@ const useDebouncedFetchSection = (
         try {
           const response = await cioClient?.autocomplete.getAutocompleteResults(
             debouncedSearchTerm,
-            autocompleteParameters
+            { ...autocompleteParameters, ...advancedParameters.autocompleteParameters }
```
deomsj, line 86: as a maintainer of this lib three months from now, I'd find this to be very confusing

instead of this:
`{ ...autocompleteParameters, ...advancedParameters.autocompleteParameters }`

I'd prefer this: 
{ ...autocompleteParameters }

deomsj (changes requested):
Nice!. This works great, I'm recommending a small change to simplify the API we expose to consumers of the lib. 

lemme know what you think, @stanlp1!

src/stories/Autocomplete/Hook/AdvancedParameters.stories.tsx
stanlp1: Good point, that makes sense to me. Updated to just combine the entire advancedParameter object with autocompleteParameters.

Our SDK should filter out any irrelevant parameters. What do you think?

https://github.com/Constructor-io/constructorio-ui-autocomplete/blob/c68b5b5531aa75fdca44086006c330972919bb80/src/hooks/useDebouncedFetchSections.ts#L63-L65

src/hooks/useDebouncedFetchSections.ts
stanlp1: 👍 Had to move some code around but this sounds good to me.

src/hooks/useDebouncedFetchSections.ts
```diff
@@ -97,7 +93,14 @@ const useDebouncedFetchSection = (
         setSectionsData({});
       }
     })();
-  }, [debouncedSearchTerm, cioClient, numTermsWithGroupSuggestions, numGroupsSuggestedPerTerm]);
+  }, [
```
deomsj, line 96: https://www.loom.com/share/7e08a6685c464e5c9dc6eb339ec17a23?sid=9bed045d-19e5-40be-8952-e39acda1de76

deomsj (changes requested):
changes are lookin great!

we have some [errors in console](https://www.loom.com/share/7e08a6685c464e5c9dc6eb339ec17a23?sid=9bed045d-19e5-40be-8952-e39acda1de76) related to hooks and state managements /  avoidable rerenders :)

src/hooks/useSections.ts
```diff
@@ -19,8 +19,12 @@ export default function useSections(
     zeroStateActiveSections ? zeroStateSections : sections
   );
 
-  const autocompleteSections = activeSections?.filter(
-    (config: UserDefinedSection) => config.type === 'autocomplete' || !config.type
+  const autocompleteSections = useMemo(
```
deomsj, line 22: nice use of the useMemo hook here!

src/hooks/useDebouncedFetchSections.ts
```diff
@@ -55,25 +55,30 @@ const useDebouncedFetchSection = (
...
+  const [autocompleteParameters, setAutocompleteParameters] =
+    useState<Partial<IAutocompleteParameters>>(advancedParameters);
   const debouncedSearchTerm = useDebounce(query);
 
-  const { numTermsWithGroupSuggestions = 0, numGroupsSuggestedPerTerm = 0 } = advancedParameters;
+  // const { numTermsWithGroupSuggestions = 0, numGroupsSuggestedPerTerm = 0 } = advancedParameters;
+  const numTermsWithGroupSuggestions = 0;
+  const numGroupsSuggestedPerTerm = 0;
 
   useEffect(() => {
```
deomsj, line 64: best to avoid managing the same state in multiple locations. Instead, when we can, its always better to derive state. More on this [here](https://dev.to/fpaghar/never-store-derived-state-heres-why-25o6) if you like.

So, rather than managing state lower in the component tree, I bet we can use the useMemo hook here as well.

Nearly everything about this will look the same, except it will be a bit easier to read the src code and there will be minor performance improvements from re-rendering perspective

so no more need for this:
`const [autocompleteParameters, setAutocompleteParameters] =
    useState<Partial<IAutocompleteParameters>>(advancedParameters);`
    
when we do something like this:
```
const decoratedAutocompleteParameters = useMemo(() => {
  const decorated = advancedParameters;
  
   if (autocompleteSections) {
        decorated.resultsPerSection = autocompleteSections.reduce(
            (acc, sectionConfig) => ({
              ...acc,
              [sectionConfig.identifier]: sectionConfig?.numResults || 8,
          }),
        {}),
      }
    }
    
  return decorated;
  }, [autocompleteSections]);
```

Disclaimer. I just wrote this code here in GH comment to demonstrate the pattern as I understand it, but some details may need fine tuning. Don't hesitate to reach out if you'd like to pair on this.

deomsj (changes requested):
Nice. This is getting better and better. 

See comments and reach out with any Qs.

esezen (commented):
This is looking good to me. Thank you both for all the great work here! We can merge after Jesse takes another look

deomsj (approved):
Excellent work on this @stanlp1!!!

PR #73, opened 2023-08-07
Error handling for tracking calls

mocca102:
Tracking calls when blocked were causing unhandled error exceptions on code sandbox
Screenshot 2023-08-07 at 7 47 26 PM

### DIFF
--- a/src/hooks/useDownShift.ts
+++ b/src/hooks/useDownShift.ts
@@ -27,17 +27,12 @@ const useDownShift: UseDownShift = ({ setQuery, items, onSubmit, cioClient, prev
         setQuery(selectedItem.value || '');
         if (selectedItem?.value) {
           if (onSubmit) onSubmit({ item: selectedItem, originalQuery: previousQuery });
-          if (!selectedItem?.data?.url) {
-            try {
+          try {
+            if (!selectedItem?.data?.url) {
               cioClient?.tracker.trackSearchSubmit(selectedItem.value, {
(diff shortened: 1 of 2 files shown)


src/hooks/useDownShift.ts
```diff
@@ -28,14 +28,24 @@ const useDownShift: UseDownShift = ({ setQuery, items, onSubmit, cioClient, prev
...
             });
+          } catch (error) {
+            // eslint-disable-next-line no-console
+            console.log(error);
           }
-          cioClient?.tracker.trackAutocompleteSelect(selectedItem.value, {
-            original_query: previousQuery,
-            section: selectedItem.section,
-          });
         }
```
jjl014, line 49: Would it make sense to combine the two try/catch statements here? For example:

```
try {
    if (!selectedItem?.data?.url) {
        cioClient?.tracker.trackSearchSubmit(selectedItem.value, {
             original_query: previousQuery,
        });
    }
    cioClient?.tracker.trackAutocompleteSelect(selectedItem.value, {
        original_query: previousQuery,
        section: selectedItem.section,
    });
} catch (error) {
    console.log(error);
}

jjl014 (approved):
LGTM! 🚀 

Do we know why the tracking calls were erroring out?

mocca102:
> LGTM! 🚀
> 
> Do we know why the tracking calls were erroring out?

I'm using Brave so it blocks tracking requests by default

PR #74, opened 2023-08-09
Add javascript usage example

mocca102:


esezen (approved):
LGTM!

esezen:
@mocca102 I noticed sections are stacked vertically on desktop. Should they be side to side instead?

PR #75, opened 2023-09-06
[CSL-2348] Add support for displaying count and images for search suggestions

esezen:


src/hooks/useCioAutocomplete.ts
```diff
@@ -117,7 +117,7 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
             onSubmit({ query });
           }
           try {
-            cioClient?.tracker.trackSearchSubmit(query, { original_query: query });
+            cioClient?.tracker.trackSearchSubmit(query, { originalQuery: query });
```
esezen, line 120: Not related to this PR but had to be updated in this PR because I updated client js version and types have been updated there

src/constants.ts
```diff
@@ -220,3 +220,5 @@ To see this in action:
 4. Next, type "short" in the example autocomplete input field.
   - Notice how the user is presented with only short sleeved items as results.
   - This is because we are filtering to the "Shirts" group`;
+
+export const termsWithImagesAndCountsDescription = `Pass boolean flags for \`displaySearchSuggestionImages\` and \`displaySearchSuggestionResultCounts\` fields to display images and counts for search suggestions.`;
```
stanlp1, line 224: Should we add a note here about making the fields displayable? So they won't just pass the flags and wonder why it doesn't work.

stanlp1 (approved):
Looks pretty good to me. Just one note about the docs.

src/constants.ts
esezen: That's a great idea. I'll update

PR #77, opened 2023-09-13
Fix failing types and add types check as github action

mocca102:


.github/workflows/check-types
```diff
@@ -0,0 +1,23 @@
+name: Run Check Types
```
esezen, line 1: This file is missing an extension. I needs to be `.yml` for it to run

esezen (approved):
LGTM!

PR #80, opened 2023-09-18
Add Remix App usage example and update Readme

mocca102:


README.md
```diff
@@ -29,7 +29,7 @@ import { CioAutocomplete } from '@constructor-io/constructorio-ui-autocomplete';
 function YourComponent() {
   return (
     <div>
-      <CioAutocomplete apiKey="key_M57QS8SMPdLdLx4x" />
+      <CioAutocomplete apiKey="key_M57QS8SMPdLdLx4x" onSubmit={(e) => {console.log(e)}} />
```
mocca102, line 32: onSubmit is required

esezen (commented):
1. Is there a way to remove the "forked from" piece?
image

2. Can we update the Readme in the Code Sandbox to say something like "This is an example for Cio Autocomplete. For more information go to XXX"

3. Is this the most lean version of the Remix app? Are there any files we can remove from the sandbox? If we can strip down the sandbox app to only required components, it might be easier for an end user to quickly understand how it's implemented

esezen:
@mocca102 not urgent, just a reminder of the above

esezen (changes requested):
Not sure why but I'm getting this and the example isn't working 🤔 
image

esezen (approved):
LGTM!

PR #83, opened 2023-09-27
[psl-2716] create a full featured and styled example

HHHindawy:


sblaurock (commented):
This looks great to me, Hossam! Let's let someone else take a closer look though as I'm not as familiar with such things 😄 

Thank you for this!

mocca102 (approved):
LGTM!

jjl014 (commented):
Nice work, this is looking great to me as well! 🚀 

I have some minor comments.

It looks like the dropdown jumps around a bit due to images loading in for the search suggestions/products. Should we give the images or their containers a `min-height` to make it feel a bit less jumpy?

Also, we're surfacing product recommendations in the zero state, but we aren't showing any images. I think it's the same way in the other example we have. Do we not show images in general for zero state recommendations? 🤔

HHHindawy:
> It looks like the dropdown jumps around a bit due to images loading in for the search suggestions/products. Should we give the images or their containers a `min-height` to make it feel a bit less jumpy?
> 

Hmmm... sure, we can do that. But I'm not sure what might be a suitable height to go with 🤔 

> Also, we're surfacing product recommendations in the zero state, but we aren't showing any images. I think it's the same way in the other example we have. Do we not show images in general for zero state recommendations? 🤔

Right now, we don't show images in zero state. This needs to be implemented. I believe we have a ticket for that. If I can't find it though I'll create one 👍 

Do you think we should remove the zero state from this example until we implement that? @jjl014

jjl014 (approved):
Nice work! Thanks for making the changes! 🚀

PR #84, opened 2023-10-03
[csl-2869] configure showing search term highlights in sections

HHHindawy:
This PR adds the option to highlight words that match the search keyword in every section the user adds the `displaySearchTermHighlights` boolean to.

Notice how the word `pant` is highlighted in all sections:

image

### DIFF
--- a/src/components/Autocomplete/SectionItem/SectionItemText.tsx
+++ b/src/components/Autocomplete/SectionItem/SectionItemText.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import { Item } from '../../../types';
+import { isInGroupSuggestion } from '../../../typeGuards';
+import { escapeRegExp } from '../../../utils';
+
+export interface SectionItemTextProps {
+  item: Item;
+  query: string;
+  highlightSearchTerm?: boolean;
+}
+
+export default function SectionItemText({
+  item,
+  query,
+  highlightSearchTerm,
+}: SectionItemTextProps) {
+  const itemText = isInGroupSuggestion(item) ? item.groupName : item.value;
+
+  if (highlightSearchTerm) {
+    const queryRegex = new RegExp(`(${escapeRegExp(query)})`, 'gi');
+    const queryIncludedInText = queryRegex.test(itemText);
+
+    if (query && queryIncludedInText) {
+      const splitText = itemText.split(queryRegex);
+
+      return (
+        <>
+          {splitText[0]}
+          <b>{splitText[1]}</b>
+          {splitText[2]}
+        </>
+      );
+    }
+  }
+
+  // eslint-disable-next-line react/jsx-no-useless-fragment
+  return <>{itemText}</>;
+}
--- a/src/components/Autocomplete/SectionItem/SectionItemText.tsx
+++ b/src/components/Autocomplete/SectionItem/SectionItemText.tsx
@@ -25,8 +25,9 @@ export default function SectionItemText({
 
       return (
         <>
-          {splitText.map((split) => {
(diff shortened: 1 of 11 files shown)


src/utils.ts
```diff
@@ -135,3 +137,33 @@ export const getActiveSectionsWithData = (
...
+  }
+
+  const regex = new RegExp(`(${escapeRegExp(query)})`, 'gi');
+
+  sections.forEach((section: UserDefinedSection) => {
+    if (section?.displaySearchTermHighlights) {
+      const sectionName = section?.displayName || section?.identifier;
+
+      document
+        .querySelector(`.${sectionName?.split(' ')?.join('.')}`)
```
stanlp1, line 157: I think generally, if possible we should try to avoid using querySelector. 

Do you think it would be possible to instead have this logic live in the SectionItem.tsx file, modify each item as they come?

That way we can work directly with an autocomplete value in React, not have to worry about sanitizing html.

stanlp1 (changes requested):
Left a comment about the current approach. Curious to see what others think.

src/utils.ts
HHHindawy: @stanlp1 Yes, that's what I did locally. I was waiting to confirm with the rest of the team, but seems like this is the right approach 👍

HHHindawy:
@stanlp1 Please have another look and let me know what you think

src/components/Autocomplete/SectionItem/SectionItemText.tsx
```diff
@@ -0,0 +1,38 @@
...
+    const queryIncludedInText = queryRegex.test(itemText);
+
+    if (query && queryIncludedInText) {
+      const splitText = itemText.split(queryRegex);
+
+      return (
+        <>
+          {splitText[0]}
+          <b>{splitText[1]}</b>
+          {splitText[2]}
```
stanlp1, line 30: There is a bug here, this will cut off the term if the query exists multiple times in the suggestion.

stanlp1 (changes requested):
Nice, the new approach looks great. There is a bug though. 

Also I am wondering if we need to use regex, instead of just splitting on the query.

src/components/Autocomplete/SectionItem/SectionItemText.tsx
HHHindawy: @stanlp1 Yes you're right, fixed that!

The reason why I'm using a regex is that we want to get all query occurrences case insensitive and replace it with the actual case that was in the original string.

I guess this is going to be tricky using the query alone.

stanlp1: 👍 That makes sense.

stanlp1 (approved):
Nice! This is looking good to me.

src/components/Autocomplete/SectionItem/SectionItemText.tsx
```diff
@@ -0,0 +1,40 @@
...
+    const queryIncludedInText = queryRegex.test(itemText);
+
+    if (query && queryIncludedInText) {
+      const splitText = itemText.split(queryRegex);
+
+      return (
+        <>
+          {splitText.map((split, index) => {
+            // eslint-disable-next-line react/no-array-index-key
+            if (queryRegex.test(split)) return <b key={`matched-word-${index}`}>{split}</b>;
```
mocca102, line 30: If you can make this an if else statement for readability it would be great. Also you can ignore this

mocca102 (approved):
LGTM! Thanks for making the changes

PR #89, opened 2023-10-11
]CSL-2874] A/b test search suggestion variants

stanlp1:


src/types.ts
```diff
@@ -27,7 +28,7 @@ export type CioAutocompleteProps = CioClientConfig & {
   openOnFocus?: boolean;
   onSubmit: OnSubmit;
   onFocus?: () => void;
-  onChange?: (string) => void;
+  onChange?: (arg0: string) => void;
```
mocca102, line 31: Can we make input or value instead of arg0

src/components/Autocomplete/SectionItem/SectionItem.tsx
```diff
@@ -11,11 +12,40 @@ export interface SectionItemProps {
   displaySearchTermHighlights?: boolean;
 }
 
+export function getSearchSuggestionFeatures(request: Partial<AutocompleteRequestType>) {
```
mocca102, line 15: I think this need to be part of the hooks logic and we only expose the features as a state returned form the hooks instead of exposing the utility function for consumers of the library to use it to get the features. And we do call `getSearchSuggestionFeatures` internally instead and expose smth like this in the hook response
```
featureToggles: {
    featureDisplaySearchSuggestionImages,
    featureDisplaySearchSuggestionResultCounts,
  }
```

@esezen what do you think?

src/utils.ts
```diff
@@ -1,6 +1,13 @@
 import ConstructorIOClient from '@constructor-io/constructorio-client-javascript';
 import { isCustomSection } from './typeGuards';
-import { OnSubmit, Item, Section, UserDefinedSection, AutocompleteResultSections } from './types';
+import {
+  OnSubmit,
+  Item,
+  Section,
+  UserDefinedSection,
+  AutocompleteResultSections,
```
mocca102, line 8: Can we remove this since it's not used anymore

src/types.ts
stanlp1: 👍 The typescript linter suggested making it arg0: string so I followed it without thinking 😂

mocca102 (commented):
Thanks for making the changes! This is looking pretty good

Just left small comments and lets get a second eye

src/components/Autocomplete/SectionItem/SectionItem.tsx
esezen: I think this makes sense. @stanlp1 anything against this approach?

stanlp1: Nope, Islam mentioned this to me, I suggested tagging you as the decision maker 😂

mocca102 (approved):
Thanks for making the changes @stanlp1 🚀

esezen (approved):
LGTM!

esezen (changes requested):
@stanlp1 I thought tests are failing on main at first but they seem to be passing there. Would you be able to check why they are failing here?

stanlp1:
> @stanlp1 I thought tests are failing on main at first but they seem to be passing there. Would you be able to check why they are failing here?

Looks like the changes I made to resolve the failing tests 3 weeks ago don't work anymore. Reverted 👍

PR #91, opened 2023-10-13
[NO-CSL] Update bundle target

esezen:


mocca102 (approved):
Thanks for making the change! LGTM!

esezen:
We will revisit this later

PR #92, opened 2023-10-19
[PARTI-505] Add GH actions based automation for publishing

KagenLH:
# PR Description
Adds some GH action based automations for publishing new versions of this package. Makes a small adjustment to the `version` npm script to accommodate these changes more cleanly (removes npm compile from this step since in the automation they are needed at different times). A detailed explanation of exactly what it should do follows:

### Update package version
* Triggered manually from the GitHub Actions UI (could also be done with scripts or GH API if desired in the future, see the [documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch)  for this trigger type)
  * Receives the version strategy to use for incrementing the new version (major, minor, patch)
  * Receives the title to use for the new release, used in the GitHub release name
* Increments the version of the package using the pre-existing script so that storybook docs, node version check, and etc still execute
* Commits the new version and pushes back to this repository
* Captures the new version of the package and the commit sha from the commit it just created for use in future steps
### Create GitHub release
* Receives the release title, new version, and commit SHA from the update package version job and creates a GitHub release using it
### Publish to NPM
* Receives the commit SHA created in the update_package_version job
* Compiles the source code for distribution
* Publishes to NPM

**All steps provide Slack notifications about completion and the full workflow is wrapped in Slack notifications indicating the start and completion**

Each step listed besides `update_package_version` is its own separate workflow and job created using the [reusable workflow system](https://docs.github.com/en/actions/using-workflows/reusing-workflows), meaning they can be individually retried on failure or triggered independently of the bigger `Publish` workflow that composes them.

## What to test
* Workflow should successfully increment the version according to the provided version strategy
* The newly incremented version and correct new commit after versioning should be used for the GitHub release and NPM publish steps
* The NPM publish job (currently done with `--dry-run` to avoid real publishing) will output the files that it would bundle in the upload to NPM. Check that those look as they should.
* Slack notifications work properly through the whole process.

### For Customer Integrations Reviewers:
* Pay special attention to the `npm version` and `npm compile` commands run here and their context, ensure that these are consistent with the way that these are expected to be used in your current manual workflow and most importantly that the crtical end outputs will be the same.

## How to test
* Visit the [fork](https://github.com/Constructor-io/parti-505-constructorio-ui-autocomplete) that this PR is based on and run the `Publish` workflow from the Actions UI

### DIFF
--- a/.github/workflows/publish-npm.yml
+++ b/.github/workflows/publish-npm.yml
@@ -1,0 +1,52 @@
+name: 'Publish: Publish to NPM'
+
+on:
+  workflow_call:
+    inputs:
+      deploy-sha:
+        description: 'SHA of the commit to deploy'
+        required: true
+        type: string
+
+  # This workflow is intended to be called from the primary publish workflow
+  # automatically. The manual trigger is maintained for partially failed workflows
+  # or edge cases that require this be run independently.
+  workflow_dispatch:
+    inputs:
+      deploy-sha:
+        description: 'SHA of the commit to deploy'
+        required: true
+        type: string
+  
+jobs:
+  publish-to-npm:
+    name: Publish to NPM
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.deploy-sha }}
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18.12.x
+          registry-url: https://registry.npmjs.org/
+          scope: '@constructorio'
+          token: ${{ secrets.NPM_TOKEN }}
+      - name: Install dependencies
+        run: npm ci
+      - name: Build compiled package for publishing
+        run: npm run compile
+      - name: Publish to NPM
+        run: npm publish --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - name: Get new release version
+        id: get-version
+        run: echo "version=v$(npm pkg get version | tr -d '\"')" >> "$GITHUB_OUTPUT"
+      - name: Notify Slack about new publish to NPM
+        uses: ravsamhq/notify-slack-action@v2
+        with:
+          notification_title: "☁️ New constructor-ui-autocomplete release published to npm at version ${{ steps.get-version.outputs.version }}"
+          status: ${{ job.status }}
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,97 @@
+name: Publish
+
+on:
+  # This allows us to run this workflow manually from the Actions tab
+  workflow_dispatch:
+    inputs:
+      version-strategy:
+        type: choice
+        description: 'Version strategy'
+        required: true
+        options:
+            - 'patch'
+            - 'minor'
+            - 'major'
+        default: patch
+      title:
+        type: string
+        description: 'Release title'
+        required: true
+
+concurrency:
+  group: deploy
+  cancel-in-progress: true
+
+# We need these permissions create new GitHub releases
+permissions: write-all
+
+jobs:
+  update_package_version:
+    name: Update package version
+    runs-on: ubuntu-latest
+    steps:
+      - name: Notify Slack about new release
+        uses: ravsamhq/notify-slack-action@v2
+        with:
+          notification_title: "⏳ Starting new constructor-ui-autocomplete deployment"
+          status: ${{ job.status }}
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+      - uses: actions/checkout@v4
+      - uses: oleksiyrudenko/gha-git-credentials@v2-latest
+        with:
+          token: '${{ secrets.GITHUB_TOKEN }}'
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '18.12.x'
+      - name: Install dependencies
+        run: npm ci
+      - name: Bump release version
+        run: npm version ${{ github.event.inputs.version-strategy }} -m "${{ github.event.inputs.title }}"
+      - name: Push main branch
+        run: git push origin main --follow-tags
+      - name: Set new release version
+        id: set-version
+        run: echo "version=v$(npm pkg get version | tr -d '\"')" >> "$GITHUB_OUTPUT"
+      - name: Set deployment SHA
+        id: set-deploy-sha
+        run: echo "deploy-sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+      - name: Notify Slack about new package version
+        uses: ravsamhq/notify-slack-action@v2
+        with:
+          notification_title: "✅ constructor-ui-autocomplete package version bumped to ${{ steps.set-version.outputs.version }} successfully"
+          status: ${{ job.status }}
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+    outputs:
+        version: ${{ steps.set-version.outputs.version }}
+        deploy-sha: ${{ steps.set-deploy-sha.outputs.deploy-sha }}
+  create_github_release:
+    name: Create GitHub release
+    needs: update_package_version
+    uses: ./.github/workflows/publish-create-github-release.yml
+    with:
+      version: ${{ needs.update_package_version.outputs.version }}
+      title: ${{ github.event.inputs.title }}
+      deploy-sha: ${{ needs.update_package_version.outputs.deploy-sha }}
+    secrets: inherit
+  publish_to_npm:
+    name: Publish to NPM
+    needs: update_package_version
+    uses: ./.github/workflows/publish-npm.yml
+    with:
+      deploy-sha: ${{ needs.update_package_version.outputs.deploy-sha }}
+    secrets: inherit
+  final_deployment_notification:
+    name: Final deployment notification
+    needs: [create_github_release, publish_to_npm]
+    runs-on: ubuntu-latest
+    if: always()
+    steps:
+      - name: Report deployment status to Slack
+        uses: ravsamhq/notify-slack-action@v2
+        with:
+          notification_title: "🏁 Constructor-ui-autocomplete deployment completed"
+          status: ${{ job.status }}
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@constructor-io/constructorio-ui-autocomplete",
-  "version": "1.9.2",
+  "version": "1.12.0",
   "description": "Constructor.io Autocomplete UI library for web applications",
   "main": "lib/cjs/index.js",
   "module": "lib/mjs/index.js",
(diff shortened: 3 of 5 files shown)


.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,48 @@
...
+          node-version: 18.12.x
+          registry-url: https://registry.npmjs.org/
+          scope: '@constructorio'
+          token: ${{ secrets.NPM_TOKEN }}
+      - name: Install dependencies
+        run: npm ci
+      - name: Build compiled package for publishing
+        run: npm run compile
+      - name: Publish to NPM
+        run: npm publish --access public
```
KagenLH, line 36: Currently using `--dry-run` on the fork repository main branch to avoid real publishes from triggering here, so don't worry about testing there making real publishes. Only this branch has it in live form.

Flagging this as a specific area of interest for the CI team to validate

.github/workflows/publish.yml
```diff
@@ -0,0 +1,100 @@
...
+      - uses: oleksiyrudenko/gha-git-credentials@v2-latest
+        with:
+          token: '${{ secrets.GITHUB_TOKEN }}'
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '18.12.x'
+      - name: Install dependencies
+        run: npm ci
+      - name: Bump release version
+        run: npm version ${{ github.event.inputs.version-strategy }}
```
KagenLH, line 50: This is the other big area I wanted to make sure we got the Customer Integrations team to look at. Should generate the new docs and increment the version in the relevant files, then push a new commit with this back to main branch for the other two jobs to consume.

package.json
```diff
@@ -35,7 +35,7 @@
     "serve-built-storybook": "npx http-server docs",
     "verify-node-version": "chmod +x ./scripts/verify-node-version.sh && ./scripts/verify-node-version.sh",
     "preversion": "node ./src/generateVersion.js",
-    "version": "npm run preversion && npm run verify-node-version && npm run build-storybook && git add -u ./docs && git add ./docs/* && git add ./src/version.ts && npm run compile",
+    "version": "npm run preversion && npm run verify-node-version && npm run build-storybook && git add -u ./docs && git add ./docs/* && git add ./src/version.ts && git add package.json",
```
KagenLH, line 38: I've adjusted this slightly for the purpose of the automation, since when we're incrementing the version  for the new commit we do not need the compiled code. The compile step will be called on its own in the publish to NPM workflow.

.github/workflows/create-github-release.yml
```diff
@@ -0,0 +1,53 @@
+name: Create GitHub release
```
pedro-lb, line 1: **Suggestion:** Should we add a prefix to this name, so that we know it's related to the publish flow?

Reason is that if you look into the repo we have a lot of workflows already, and they're sorted alphabetically. It would be a bit easier to understand what's related to the publish flow if it was all grouped together.

```suggestion
name: Publish: Create GitHub release
```

Note: do not forget to rename the file (to add the prefix too) in case we adopt this.

image

.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,48 @@
+name: Publish to NPM
```
pedro-lb, line 1: **Suggestion:** Should we add a prefix to this name, so that we know it's related to the publish flow?

Reason is that if you look into the repo we have a lot of workflows already, and they're sorted alphabetically. It would be a bit easier to understand what's related to the publish flow if it was all grouped together.

```suggestion
name: Publish: Publish to NPM
```

Note: do not forget to rename the file (to add the prefix too) in case we adopt this.

image

.github/workflows/create-github-release.yml
```diff
@@ -0,0 +1,53 @@
...
+        with:
+          name: ${{ inputs.title }}
+          tag: ${{ inputs.version }}
+          token: ${{ github.token }}
+          generateReleaseNotes: true
+          commit: ${{ inputs.deploy-sha}}
+      - name: Notify Slack about successful release
+        uses: ravsamhq/notify-slack-action@v2
+        with:
+          notification_title: "New constructor-ui-autocomplete GitHub release created at version ${{ inputs.version }}"
```
pedro-lb, line 50: **Small suggestion:** Every other notification has emojis, should we add one here too?

```suggestion
          notification_title: "😸 New constructor-ui-autocomplete GitHub release created at version ${{ inputs.version }}"
```

.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: npm publish --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+      - name: Get new release version
+        id: get-version
+        run: echo "version=$(npm pkg get version)" >> "$GITHUB_OUTPUT"
+      - name: Notify Slack about new publish to NPM
+        uses: ravsamhq/notify-slack-action@v2
+        with:
+          notification_title: "New constructor-ui-autocomplete release published to npm at version ${{ steps.get-version.outputs.version }}"
```
pedro-lb, line 45: **Small suggestion:** Every other notification has emojis, should we add one here too?

```suggestion
          notification_title: "☁️ New constructor-ui-autocomplete release published to npm at version ${{ steps.get-version.outputs.version }}"
```

package.json
pedro-lb, line 38: **Issue:** Whenever we bump the `package.json` version, it will make auto generated changes to `package-lock.json` whenever installing dependencies with `npm i`. I think we should also commit those auto generated changes.

For example, after manually updating package.json and running `npm i`:

image

So, I think we should:

- Run `npm i` to update the lock
- Also `git add package-lock.json`

pedro-lb (changes requested):
Seriously awesome work here so far @KagenLH! The implementation is super clean with reusable workflows and I think this is going to be a major win to automate the process across multiple repos.

I left a few suggestions, with one that is slightly a blocker - but I think we'll be able to ship it after solving those!

package.json
KagenLH: Oooh this is a real issue great catch will get this fixed up

.github/workflows/create-github-release.yml
```diff
@@ -0,0 +1,53 @@
...
+jobs:
+  create-github-release:
+    name: Create GitHub release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create release
+        uses: ncipollo/release-action@v1
+        with:
+          name: ${{ inputs.title }}
+          tag: ${{ inputs.version }}
```
esezen, line 43: It appears like this produces a version like `"1.10.0"`
image
We usually do `v1.10.0`
image

Can we strip the `"` and add a `v` in front? Alternatively, is there a way to infer this from package.json?

.github/workflows/publish.yml
esezen, line 50: Should this be something like `npm version ${{ github.event.inputs.version-strategy }} -m ${{ github.event.inputs.title }}`?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,100 @@
...
+      - name: Push main branch
```
esezen, line 51: When we are pushing up, we need `git push main --follow-tags` can we add `--follow-tags` to this or does it do that automatically?

package.json
esezen, line 38: npm version command should handle adding package and package-lock itself. We shouldn't be manually adding those. They should be staged and included in the commit. Was it not adding these files when you tested?

.github/workflows/create-github-release.yml
esezen: I think I found the answer to my question. It looks like value gets pushed to this from the previous step. Can we still add the `v` and get rid of `"`"

esezen (commented):
This is super duper cool! 🚀 
I left a few comments here and there but the overall process is looking great

.github/workflows/create-github-release.yml
pedro-lb: Great catch!

.github/workflows/create-github-release.yml
KagenLH: I think this is good, changed in the latest commit

.github/workflows/publish-npm.yml
KagenLH: I've changed the name of the job in the file for grouping, but I didn't change the file name since it already starts with publish and I think it would get weird to double it. Let me know if you think that we should still change the name for this one.

.github/workflows/create-github-release.yml
KagenLH: Added in the latest commit!

.github/workflows/publish-npm.yml
KagenLH: Added in the latest commit!

package.json
KagenLH: Turns out this is actually already handled by the `npm version` command along with the `package.json`. I removed this addition in response to the comment from @esezen

.github/workflows/create-github-release.yml
KagenLH: Made this adjustment with the latest commit!

.github/workflows/publish.yml
KagenLH: You're right. I added this with the latest commit

.github/workflows/publish.yml
KagenLH: This action doesn't support it, but for this I think we can simply run the command. I've changed that in the most recent commit.

package.json
KagenLH: You're totally right. I removed this in the most recent commit.

.github/workflows/publish-npm.yml
deomsj: very thoughtful setup for testing, thanks for calling this out @KagenLH  :)

.github/workflows/publish.yml
deomsj: ah cool. Suppose we are safe to test this functionality over on [the forked repo](https://github.com/Constructor-io/parti-505-constructorio-ui-autocomplete), ya?

package.json
deomsj: I see that the previous and edited version of this "version" command are explictly calling `npm run preversion` ...

I don't think we need to call pre and post commands explicitly. [npm docs on topic here](https://docs.npmjs.com/cli/v10/using-npm/scripts#pre--post-scripts)

deomsj: actually, I'm wondering if we want to refactor this:
```
    "preversion": "node ./src/generateVersion.js",
    "version": "npm run preversion && npm run verify-node-version && npm run build-storybook && git add -u ./docs && git add ./docs/* && git add ./src/version.ts",
```

to something like this:
```
    "version": "node ./src/generateVersion.js && npm run verify-node-version",
     "postversion": "npm run build-storybook && git add -u :/docs && git add ./src/version.ts",
```

deomsj (changes requested):
Awesome work on this, @KagenLH!!! 

I learned a lot reviewing the code and supporting descriptions in this PR. Very cool stuff. 

I left a few questions/idea as inline comments and I'm looking forward to reviewing live together shortly.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@constructor-io/constructorio-ui-autocomplete",
-  "version": "1.9.2",
```
KagenLH, line 3: The diff seems to have broken somehow and it thinks that upstream main is on 1.9.2... Should not be any problem when we merge it since its actually on 1.9.3 currently.

.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,48 @@
+name: 'Publish: Publish to NPM'
+
+on:
+  workflow_dispatch:
```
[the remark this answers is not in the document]
KagenLH: Rearranged and added some explanatory comments to the workflow file in the newest version, as per discussion in our sync. For the context of others: maintaining a manual trigger for the sub-workflow is useful in case something goes wrong with one of them and they need to be re-triggered with new endpoints, or there is an edge case where we want to run an npm publish or github release on its own.

package.json
KagenLH: I've removed the unneeded preversion call, but as per our sync discussion we can't do the docs generation and git staging in our postversion since the commit is already created and tagged by this point.

deomsj (approved):
🟢 

Awesome work here @KagenLH!!! 

Let me know if there is anything I can do to help move this forward. As far as I can tell this is ready to ship!

pedro-lb (approved):
So far it's looking good to me! Let's handle all other feedback and we should be good to :shipit:!

jjl014 (approved):
This looks great from what I can tell! 🚀 Fantastic job! Really appreciate the detailed description and comments as well, it was super helpful! 

Also, thank you @deomsj @pedro-lb @esezen for giving this a thorough review! ❤️

PR #96, opened 2023-11-03
Populate autocomplete input with selected term only for Search Suggestions

mocca102:


src/hooks/useDownShift.ts
```diff
@@ -24,7 +24,11 @@ const useDownShift: UseDownShift = ({ setQuery, items, onSubmit, cioClient, prev
     itemToString: (item) => item?.value || '',
     onSelectedItemChange({ selectedItem }) {
       if (selectedItem) {
-        setQuery(selectedItem.value || '');
+        if (selectedItem?.section === 'Search Suggestions') {
+          setQuery(selectedItem.value);
+        } else {
+          setQuery('');
+        }
```
guiquintelas, line 31: **Suggestion**: I think it's better to keep the search input as it is, without cleaning it.

If the customer wants to add a special behavior, it's always possible to add an on-click listener. But it would be impossible to remove this "clear input" behavior.

```suggestion
        if (selectedItem?.section === 'Search Suggestions') {
          setQuery(selectedItem.value);
        }
```

guiquintelas (changes requested):
Thanks for the quick fix Islam! It's working great 👌 


https://github.com/Constructor-io/constructorio-ui-autocomplete/assets/29166076/3bbb73f2-74f7-496e-a35a-f0ab81d5479d

_ignore the abrupt style changes_ 

After the little change is implemented we can 🚢  it!

guiquintelas:
We also have some failing checks ... Not sure how those are usually handled.

mocca102:
> We also have some failing checks ... Not sure how those are usually handled.

We have a separate PR to handle the flaky failing tests

src/hooks/useDownShift.ts
mocca102: I like that

src/hooks/useDownShift.ts
```diff
@@ -24,7 +24,9 @@ const useDownShift: UseDownShift = ({ setQuery, items, onSubmit, cioClient, prev
     itemToString: (item) => item?.value || '',
     onSelectedItemChange({ selectedItem }) {
       if (selectedItem) {
-        setQuery(selectedItem.value || '');
+        if (selectedItem?.section === 'Search Suggestions') {
+          setQuery(selectedItem.value || '');
```
mocca102, line 28: Added ` || ''` because I believe it handled an edge case

PR #98, opened 2023-11-06
[CSL-2971] Update dependencies

esezen:


jjl014 (approved):
Whoa, 15,312 additions and 3,798 deletions! 🤯 

LGTM! 🚀 Looks like there were some failing tests, but they're also failing on main.

PR #101, opened 2023-11-13
[NO-CSL] Update tests to use another term

esezen:


mocca102 (approved):
Passing for me as well. Thank you!

PR #102, opened 2023-11-13
[CSL-3029] Update isProduct type guard

esezen:
The `isProduct` type guard was broken in [this PR](https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/75/files#diff-9aea99b367f9871cb2e4c1a75340c75c63d6777954d9f882d06417d192d3894b). That PR incorrectly assumes anything outside of the `Products` section would be a non-product but that assumption is not correct. Recommendations would have a different section and could have an image. This PR reverts the type guard but also makes sure anything in `Search Suggestions` section will return false.

### DIFF
--- a/src/hooks/useFetchRecommendationPod.ts
+++ b/src/hooks/useFetchRecommendationPod.ts
@@ -19,13 +19,13 @@ const useFetchRecommendationPod = (
       );
       const recommendationPodResults = {};
 
-      responses.forEach(({ response }) => {
+      responses.forEach(({ response }, index) => {
         const { pod, results } = response;
         if (pod?.id) {
           recommendationPodResults[pod.id] = results?.map((item: Item) => ({
             ...item,
             id: item?.data?.id,
-            section: pod.id,
+            section: recommendationPods[index]?.section || 'Products',
           }));
         }
       });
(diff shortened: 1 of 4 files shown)


src/typeGuards.ts
```diff
@@ -8,7 +8,7 @@ import {
 } from './types';
 
 export function isProduct(item: Item): item is Product {
-  return (item as Product).section === 'Products';
+  return item.section !== 'Search Suggestions' && (item as Product).data?.image_url !== undefined;
```
mocca102, line 11: What do you think about repurposing isSearchSuggestion utility here instead of `item.section !== 'Search Suggestions' ` to be `!isSearchSuggestion(item)`

src/hooks/useFetchRecommendationPod.ts
```diff
@@ -19,13 +19,13 @@ const useFetchRecommendationPod = (
...
       const recommendationPodResults = {};
 
-      responses.forEach(({ response }) => {
+      responses.forEach(({ response }, index) => {
         const { pod, results } = response;
         if (pod?.id) {
           recommendationPodResults[pod.id] = results?.map((item: Item) => ({
             ...item,
             id: item?.data?.id,
-            section: pod.id,
```
jjl014, line 28: Are we using the pod id for anything? If so, we might want to keep it around

jjl014 (approved):
This looks great to me! 🚀 

Just left one small comment.

src/hooks/useFetchRecommendationPod.ts
esezen: As far as I can tell, we don't use it anywhere

esezen:
@mocca102 Can you give this another review?

mocca102 (approved):
LGTM!

PR #103, opened 2023-11-17
[PARTI-542] Autocomplete UI Library: Handle permissions on protected branches and implement `npm version` already existing logic

guiquintelas:
## Changes
- Changes both `package.json` and `package-lock.json` version fields to `0.0.0`.
- Update both `publish.yml` and `publish-npm.yml` workflows to track the library's version using git tags and releases.
- Change the `git push` to only push the latest created tag.
- Fix final Slack notification wrong status.
- Adds a `Deploy Storybook` workflow.
- Adds a `Publish to CDN` workflow.
- Changes `publish.yml` to:
  - Also deploy all storybook changes to GH pages.
  - Updates the `version.ts` file before publishing the new NPM version.

Some [context](https://constructor.slack.com/archives/C04B29BNW9H/p1699374608096719) regarding the `0.0.0` version change.

Feel free to test these workflow changes using this [fork](https://github.com/Constructor-io/parti-505-constructorio-ui-autocomplete/actions/workflows/publish.yml).


README.md
```diff
@@ -202,6 +202,23 @@ npm run compile           # generate lib folder for publishing to npm
...
+- **Version Strategy**: `major`, `minor`, or `patch`.
+- **Title**: A title for the release.
+
+This workflow will automatically:
+1. Bump the library version using the provided strategy.
+2. Create a new git tag.
+3. Create a new GitHub release.
+4. Compile the library.
+5. Publish the new version to NPM.
+6. Report the progress on the [#integrations-os-ui-deployment-notifications](https://constructor.slack.com/archives/C061D3CFVR9) Slack channel.
```
pedro-lb, line 217: **Suggestion:** The repo is public, so let's remove the slack channel name

```suggestion
6. Report the progress on the [relevant Slack channel](https://constructor.slack.com/archives/C061D3CFVR9).
```

.github/workflows/publish.yml
```diff
@@ -44,18 +44,51 @@ jobs:
...
+        id: last_release
+        uses: InsonusK/get-latest-release@v1.0.1
+        with:
+          myToken: ${{ github.token }}
+          view_top: 1
+
+      - name: Prepare version and strategy arguments
+        id: prepare-args
+        run: |
+            # remove the "v" preffix.
```
pedro-lb, line 61: **Typo**

```suggestion
            # remove the "v" prefix.
```

.github/workflows/helpers/map-strategy.js
```diff
@@ -0,0 +1,19 @@
...
+    outputString = "feature";
+    break;
+  case "patch":
+    outputString = "bug";
+    break;
+  default:
+    outputString = "unknown";
+}
+
+console.log(outputString);
```
pedro-lb, line 19: **Suggestion:** Instead of having this file, I think we should simply have the same values in the dropdown - so that we don't need to map anything in a js script.

Think that this workflow will be reused in other places, so the simpler it is the better it is to reuse 🥳

pedro-lb (changes requested):
Awesome work here @guiquintelas! 🔥🔥 

Left some reviews, I think we'll be good to go once those are solved!

README.md
guiquintelas: That's right! Nice catch!

.github/workflows/publish-npm.yml
```diff
@@ -26,27 +26,49 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.deploy-sha }}
```
pedro-lb, line 28: **Issue:** we can drop this.

.github/workflows/publish-npm.yml
```diff
@@ -26,27 +26,49 @@ jobs:
...
+          fetch-tags: true
+          # fetch all history so that git describe works
+          fetch-depth: 0
```
pedro-lb, line 31: **Suggestion:** We can drop this if we're receiving the version as an input.

Let's receive the version in the input and use it here

pedro-lb (changes requested):

.github/workflows/publish-npm.yml
pedro-lb: Check this pr for an implementation suggestion:

- https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/104/files

.github/workflows/helpers/map-strategy.js
pedro-lb: Note: we can use another action that maps 1:1 as discussed

guiquintelas: Done! But sorry about the huge commit. I was continuously testing on a fork branch, so I just moved over all the changes at once 😬

.github/workflows/publish-npm.yml
guiquintelas: Done! ✅

.github/workflows/publish.yml
```diff
@@ -22,76 +22,100 @@ concurrency:
   group: deploy
   cancel-in-progress: true
 
-# We need these permissions create new GitHub releases
-permissions: write-all
-
 jobs:
+  deploy_storybook:
+    name: Deploy Storybook
+    uses: ./.github/workflows/deploy-storybook.yml
```
pedro-lb, line 28: **Suggestion:** Should we deploy the storybook after publishing to npm and cdn? 

That way if something fails we'd avoid updating it

package.json
```diff
@@ -13,10 +13,6 @@
       "import": "./lib/mjs/index.js",
       "require": "./lib/cjs/index.js"
     },
-    "./cjs": {
-      "import": "./lib/cjs/index.js",
-      "require": "./lib/cjs/index.js"
-    },
```
pedro-lb, line 19: **Question:** should we remove this?

pedro-lb (commented):
Awesome work here @guiquintelas! 🔥🔥🔥🔥

Left a few comments, we should be good to go after solving those! 💯

.github/actions/create-version-tag/action.yml
```diff
@@ -0,0 +1,58 @@
...
+      run: |
+        echo "version=v$(jq '.version' package.json --raw-output)" >> "$GITHUB_OUTPUT"
+
+    - name: Create release git tag
+      shell: bash
+      run: git tag -fa ${{ steps.next-version.outputs.version }} -m "${{ github.event.inputs.title }}"
+
+    - name: Push tag to main branch
+      shell: bash
+      # We use --force here to handle any already existing tags - In case we need to re-run this workflow.
```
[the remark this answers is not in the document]
guiquintelas: I thought about that. I tried to minimize possible developer errors that might break this workflow. In this case, the developer would forget about the new workflow and locally run a npm version command. This would guarantee that any wrongly created tag won't break this workflow.

Another upside I see with this approach is the ease of use to just re-run the main `Publish` workflow, in case the first step fails, instead of reaching into manual steps for everything else. The tag version is always based on the last release made, so if we're trying to create the same tag again, it's definitely because something went wrong, and we're trying to fix it.

But I might be missing some other edge cases that we might not want to use the `--force` flag. Wdyt?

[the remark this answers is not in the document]
pedro-lb: That makes sense! I think we're fine to use it. Let's leave it as is ✅

pedro-lb (approved):
LGTM! Awesome, awesome work here @guiquintelas! 🔥🔥🔥🔥

I think we're good to go!

.github/workflows/deploy-storybook.yml
```diff
@@ -0,0 +1,53 @@
...
+        uses: actions/configure-pages@v2
+
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v1
+        with:
+          path: './docs'
+
+      - name: Deploy to GitHub Pages
+        id: deployment
+        uses: actions/deploy-pages@v1
```
esezen, line 53: Would you be able to explain how this works? Our docs (GH Pages) are served from `./docs` on main. How does this action deploy docs without having write access to main?

guiquintelas: For sure! 



GH provides two ways to _deploy_ using GH pages:
1. **From a branch**: Directly consumes all files right from the branch content - given the correct path.
2. **From GH Actions**: We're required to manually upload our files to GH.



We can perform this upload using the `actions/upload-pages-artifact` action. As described in this [GitHub article](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow).

You can also see a working example over on [this repo](https://github.com/Constructor-io/connect-template-toolkit/blob/main/.github/workflows/deploy-docs-to-gh-pages.yml).

Finally, answering your question. We simply don't need to change any files on the `main` branch, GH stores all resulting built files as artifacts directly consumed by GH Pages.

.github/actions/create-version-tag/action.yml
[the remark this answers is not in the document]
esezen: Thanks for the explanation @guiquintelas !

> In this case, the developer would forget about the new workflow and locally run a npm version command.

They would have to push the created tag to main as well, right? And wouldn't the job override the existing tag in this case? In a way, it would be failing silently. There was something wrong with the tag, but the job fixed it without producing an error.

Would it make more sense to fail in this case? First of all, no one should push a tag to main. If they do, they need to remove it when they are aware (either on GH or locally via git). I think our git tags should always be "clean" and when something is wrong like a developer pushing a tag accidentally and running the job afterwards, I think the job should fail, explicitly call out the error, and someone needs to go to tags and clean it up before running the job again. 

-----------

Another scenario is, if I run the job with an existing version, would that mean it will override that tag? Again I think it might be safer to error out in this case. If someone is trying to push a version that already exists on NPM or in our git tags, that shouldn't be allowed or we would be re-writing history. 

(Let me know if I'm completely misunderstanding this)

.github/actions/create-version-tag/action.yml
```diff
@@ -0,0 +1,58 @@
...
+      run: git push --force origin ${{ steps.next-version.outputs.version }}
```
esezen, line 50: This will only push the tag and nothing else, right?

.github/workflows/create-github-release.yml
```diff
@@ -29,26 +28,22 @@ on:
...
-        required: true
 
 jobs:
   create-github-release:
     name: Create GitHub release
     runs-on: ubuntu-latest
     steps:
       - name: Create release
-        uses: ncipollo/release-action@v1
+        uses: ncipollo/release-action@6c75be85e571768fa31b40abf38de58ba0397db5 # v1.13.0
```
esezen, line 38: Just out of curiosity, why are we not using the version number here instead of the commit?

.github/workflows/publish.yml
```diff
@@ -22,76 +22,104 @@ concurrency:
...
     name: Publish to NPM
     needs: update_package_version
     uses: ./.github/workflows/publish-npm.yml
     with:
-      deploy-sha: ${{ needs.update_package_version.outputs.deploy-sha }}
+      version: ${{ needs.update_package_version.outputs.version }}
     secrets: inherit
+  
+  # disabled until we fully test it
+  publish_to_cdn:
```
esezen, line 87: Would the workflow run this step if `publish_to_npm` fails?

.github/workflows/publish-cdn.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@010d0da01d0b5a38af31e9c3470dbfdabdecca3a # v4.0.1
+        with:
+          role-session-name: cdn-publish-action
+          role-to-assume: ${{ env.GH_ACTION_ROLE }}
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Copy files to CDN
+        run: |
+          cp dist/constructorio-ui-autocomplete-bundled dist/${{ inputs.version }}.js
```
esezen, line 53: Does the first argument need `.js` extension? How does it work without the extension 🤔

esezen (commented):
This is amazing @guiquintelas ! I love all the `jq` wizardry 🧙 
I left some comments, they are mostly me trying to understand things. Thanks for working on this, super excited to get this merged after BF 🚀🚀🚀

.github/actions/compile/action.yml
```diff
@@ -0,0 +1,23 @@
...
+inputs:
+  version: 
+    description: 'The current version.'
+    type: string
+    required: true
+    
+runs:
+  using: "composite"
+  steps:
+    - name: Genarate new src/version.ts file
```
pedro-lb, line 14: **Typo**

```suggestion
    - name: Generate new src/version.ts file
```

.github/workflows/publish-cdn.yml
pedro-lb: It does! My bad - forgot to add it.

Fixed in 10e4f3e6fb7a1555e27ea7869418f18547be0951.

.github/actions/compile/action.yml
```diff
@@ -0,0 +1,23 @@
+name: Compile Code
+
+description: 'Creates all code artefacts required to publish a new version.'
```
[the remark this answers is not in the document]
pedro-lb: Fixed during the merge with my pr: f1dc3a43a4d91a86f57ff52ecf5d4fae90d447a0

.github/actions/compile/action.yml
pedro-lb: Fixed during the merge with my pr: f1dc3a43a4d91a86f57ff52ecf5d4fae90d447a0

.github/workflows/publish.yml
```diff
@@ -22,76 +22,104 @@ concurrency:
...
   publish_to_npm:
     name: Publish to NPM
     needs: update_package_version
     uses: ./.github/workflows/publish-npm.yml
     with:
-      deploy-sha: ${{ needs.update_package_version.outputs.deploy-sha }}
+      version: ${{ needs.update_package_version.outputs.version }}
     secrets: inherit
+  
+  # disabled until we fully test it
```
pedro-lb, line 86: **Block:** Need to re-enable it once the CDN tests are done

.github/workflows/publish-cdn.yml
```diff
@@ -53,8 +53,8 @@ jobs:
 
       - name: Copy files to CDN
         run: |
-          cp dist/constructorio-ui-autocomplete-bundled dist/${{ inputs.version }}.js
-          aws s3 sync dist/${{ inputs.version }}.js s3://constructor-frontend-dev/ui/autocomplete --cache-control 'public, max-age=1800' --acl 'public-read'
+          cp dist/constructorio-ui-autocomplete-bundled.js dist/${{ inputs.version }}.js
+          aws s3 sync dist/${{ inputs.version }}.js s3://constructor-frontend-dev/ui/autocomplete/test --cache-control 'public, max-age=1800' --acl 'public-read'
```
pedro-lb, line 57: **Block:** make sure to get the correct s3 path + extension

pedro-lb (changes requested):
Some things to remember on the cdn front

.github/workflows/publish.yml
pedro-lb: 43fc868fed28aa04d90d87d74d0c706df8b28c31

.github/workflows/publish-cdn.yml
pedro-lb: It's correct now

pedro-lb:
Publish to CDN is working as expected:

image

image

.github/workflows/publish-cdn.yml
```diff
@@ -52,9 +52,19 @@ jobs:
           aws-region: us-east-1
 
       - name: Copy files to CDN
+        shell: bash
         run: |
-          cp dist/constructorio-ui-autocomplete-bundled dist/${{ inputs.version }}.js
-          aws s3 sync dist/${{ inputs.version }}.js s3://constructor-frontend-dev/ui/autocomplete --cache-control 'public, max-age=1800' --acl 'public-read'
+          # remove the "v" prefix from the version
+          current_version=$(echo "${{ inputs.version }}" | tr -d 'v')
```
pedro-lb, line 58: **Suggestion**: @guiquintelas small suggestion here, but maybe we could have an action (e.g. just like the `compile` one) to build this and throw it into the `env`?

I can see that we're doing the same thing in 3 places already, so it might be a bit flaky to maintain

pedro-lb (commented):
Some more comments - we should be good to go after solving the pending ones!

README.md
```diff
@@ -195,12 +195,31 @@ npm run dev             # start a local dev server for Storybook
 npm run lint            # run linter
 ```
 
-### Maintain Library
+## Publishing new versions
+
+Dispatch the [Publish](https://github.com/Constructor-io/constructorio-ui-autocomplete/actions/workflows/publish.yml) workflow in GitHub Actions. You're required to provide two required arguments:
```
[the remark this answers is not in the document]
guiquintelas: Nice catch! Typical copilot mistake 🤦 

194cee5f2d9f8c0b07a1e48245e9231ea4129291 Done ✅

.github/actions/compile/action.yml
```diff
@@ -0,0 +1,23 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Genarate new src/version.ts file
+      shell: bash
+      run: |
+        # remove the "v" preffix.
+        current_version=$(echo "${{ inputs.version }}" | tr -d 'v')
+
+        echo "export default '$current_version';" > src/version.ts
```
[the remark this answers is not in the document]
guiquintelas: A 100% safe way, I'm not so sure. However, I made a test commit that performed a `cat` command after compiling the library. So we could see the resulting `.js` file created.

image

[Here is the full workflow execution](https://github.com/Constructor-io/parti-505-constructorio-ui-autocomplete/actions/runs/6936259743/job/18868096118).

.github/actions/create-version-tag/action.yml
guiquintelas: Yupp, that's right! 

[Source](https://stackoverflow.com/a/5195913/10745645)

.github/workflows/create-github-release.yml
guiquintelas: A new security best practice that I found out on the platform infrastructure's _GitHub Actions Security Cheat Sheet_ article. 

> When using third-party actions, pin the version with a commit hash rather than a tag to shield your workflow from potential supply-chain compromise:

> Tags can be moved or deleted, repeating SHA requires SHA-1 collision for valid Git object payload.

.github/workflows/publish.yml
guiquintelas: It would. The reason was to streamline the happy path, making it as fast as possible. But the sad path would increase its troubleshooting complexity. 

I'm not sure if the tradeoffs are worth it. We would have a longer publishing process, but I'm pretty sure that's not so time-sensitive. But I know that any surprise complexity trying to debug a workflow error will only make the sad path sadder.

Done b6f94df5ad01006b6570d6e71ef639658aa93b6d ✅ 

@pedro-lb any concerns about running all actions sequentially?

.github/actions/create-version-tag/action.yml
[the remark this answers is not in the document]
guiquintelas: Totally agree @esezen.

I had the wrong mindset that the developer would just re-run the whole pipeline when encountering an error. That could cause many cascading errors. The best way is to run each subworkflow individually. And the best way to ensure that is to prevent the developer from re-running the main pipeline (supposing the error happened after the tag creation - most likely).

I imagine the two most probable actions to fail are:
1. Compiling the library.
2. Compiling the storybook.

The course of action would then be fixing the error and:
1. Manually re-running:
  a. Publish to NPM.
  b. Publish to CDN.
  c. Deploy storybook (if needed).
2. Just re-run the "Deploy storybook" action.

In both cases, we wouldn't be able to re-run the main pipeline because the GH release would have already been created. 

> I think the job should fail, explicitly call out the error, and someone needs to go to tags and clean it up before running the job again.

This 👌 

Thanks for shining a light on this @esezen!

.github/workflows/publish.yml
pedro-lb: No concerns at all. I think it's easier to manage if we run it sequentially.

.github/workflows/publish-cdn.yml
guiquintelas: I tried implementing with a shared `env`, but we would lose our failsafe when manually running jobs. If anyone accidentally fills in the version with a prefix, we can safely handle it.

So I went ahead and created a new action just for it.

Done 6ce74b6017d7ff9b47ea0256031a4d69cffc8bae ✅

.github/actions/remove-version-prefix/action.yml
```diff
@@ -0,0 +1,22 @@
...
+  using: "composite"
+  steps:
+    - name: Remove it
+      shell: bash
+      id: action
+      run: |
+        # remove the "v" prefix.
+        version=$(echo "${{ inputs.version }}" | tr -d 'v')
+
+        # update the package.json file with the current version.
```
pedro-lb, line 21: **Issue:** I think this comment is wrong.

Maybe we can update it to something like `Adds the version to output`

pedro-lb (approved):
LGTM! I left a minor discussion but I think we're ready to go after solving it.

Awesome work here, @guiquintelas! 🚀 🔥

.github/actions/remove-version-prefix/action.yml
guiquintelas: Oops, done c13cd57dcd8c858cb3007278eba6058edaf06394

pedro-lb (approved):
LGTM!

PR #104, opened 2023-11-20
[PARTI-515] Add action to deploy to CDN

pedro-lb:
# PR Description

This PR adds the following changes:

- Adds a new action to deploy to the CDN
- Triggers this action on the Publish workflow

## ⚠️ Dependencies

- https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/103

### DIFF
--- a/.github/workflows/publish-cdn.yml
+++ b/.github/workflows/publish-cdn.yml
@@ -48,7 +48,8 @@ jobs:
 
       - name: Copy files to CDN
         run: |
-          aws s3 sync ./lib/cjs s3://constructor-frontend-dev/ui/autocomplete --cache-control 'public, max-age=1800' --acl 'public-read'
+          cp dist/constructorio-ui-autocomplete-bundled dist/${{ inputs.version }}.js
+          aws s3 sync dist/${{ inputs.version }}.js s3://constructor-frontend-dev/ui/autocomplete --cache-control 'public, max-age=1800' --acl 'public-read'
 
       - name: Notify Slack about new publish to CDN
         uses: ravsamhq/notify-slack-action@bca2d7f5660b833a27bda4f6b8bef389ebfefd25 # v2.3.0
--- a/.github/workflows/publish-cdn.yml
+++ b/.github/workflows/publish-cdn.yml
@@ -40,7 +40,7 @@ jobs:
         run: npm run compile
 
       - name: Configure AWS credentials
-        uses: aws-actions/configure-aws-credentials@v4
+        uses: aws-actions/configure-aws-credentials@010d0da01d0b5a38af31e9c3470dbfdabdecca3a # v4.0.1
         with:
           role-session-name: cdn-publish-action
           role-to-assume: ${{ env.GH_ACTION_ROLE }}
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-cdn.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-session-name: cdn-publish-action
+          role-to-assume: ${{ env.GH_ACTION_ROLE }}
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Copy files to CDN
+        run: |
+          aws s3 sync ./lib/cjs s3://constructor-frontend-dev/ui/autocomplete --cache-control 'public, max-age=1800' --acl 'public-read'
```
guiquintelas, line 51: **Issue**: I think we're uploading the wrong file. 

We need to upload the `dist/constructorio-ui-autocomplete-bundled.js` instead. That's the only file needed for bundled usage.

.github/workflows/publish-cdn.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Notify Slack about new publish to CDN
+        uses: ravsamhq/notify-slack-action@v2
```
guiquintelas, line 54: **Suggestion**: Let's use fixed commit sha's with third-party actions.

```suggestion
        uses: ravsamhq/notify-slack-action@bca2d7f5660b833a27bda4f6b8bef389ebfefd25 # v2.3.0
```

.github/workflows/publish-cdn.yml
```diff
@@ -0,0 +1,59 @@
...
+          node-version: 18.12.x
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build compiled package for publishing
+        run: npm run compile
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
```
guiquintelas, line 43: **Suggestion**: We should use a fixed commit sha here.

guiquintelas (changes requested):
Looking good Pedro! Just found a minor issue that we need to solve. And left a couple of other suggestions

.github/workflows/publish-cdn.yml
pedro-lb: Done! 8db92ca7dca340405e2a7d5ec6c44faa0d424e99

.github/workflows/publish-cdn.yml
pedro-lb: Done! 8db92ca7dca340405e2a7d5ec6c44faa0d424e99

.github/workflows/publish-cdn.yml
pedro-lb: Oh, great catch! I also renamed the file to upload it to the proper path (e.g. `/v1.0.0.js`)

845044097dd3c0983f73eccf1a47b12bf1196a1f

guiquintelas (approved):
Ready to be shipped!

PR #105, opened 2023-11-22
[NO-REF] Add `publish-cdn` workflow

pedro-lb:
## Changes

- Adds the publish to CDN workflow from [this pr](https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/103)


guiquintelas (approved):
Approving just so this workflow could be tested over on https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/105

PR #108, opened 2023-11-27
[CSL-3049] Add support for `cioJsClientOptions`

esezen:
Decided to call the parameter `cioJsCientOptions` to match the [types](https://github.com/Constructor-io/constructorio-client-javascript/blob/master/src/types/index.d.ts#L40).

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -111,6 +111,7 @@ export const apiKeyDescription = `Pass an \`apiKey\` to request results from con
 export const cioJsClientDescription = `If you are already using an instance of the \`ConstructorIOClient\`, you can pass a \`cioJsClient\` instead of an \`apiKey\` to request results from constructor's servers
 
 > Note: when we say \`cioJsClient\`, we are referring to an instance of the [constructorio-client-javascript](https://www.npmjs.com/package/@constructor-io/constructorio-client-javascript)`;
+export const cioJsClientOptionsDescription = `If you don't want to create an instance of the \`ConstructorIOClient\` but still want to customize some of the options, you can pass a \`cioJsClientOptions\` object. You can learn more about the possible values [here under the parameters section](https://constructor-io.github.io/constructorio-client-javascript/ConstructorIO.html).`;
 export const placeholderDescription = `Pass a \`placeholder\` to override the default input field placeholder text shown to users before they start typing their query`;
 export const searchSuggestionsDescription = `Override default \`sections\` to only suggest search terms`;
 export const productsDescription = `Override default \`sections\` to only suggested products`;
--- a/src/hooks/useCioAutocomplete.ts
+++ b/src/hooks/useCioAutocomplete.ts
@@ -28,6 +28,7 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
     openOnFocus,
     apiKey,
     cioJsClient,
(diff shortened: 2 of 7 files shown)


src/constants.ts
```diff
@@ -111,6 +111,7 @@ export const apiKeyDescription = `Pass an \`apiKey\` to request results from con
 export const cioJsClientDescription = `If you are already using an instance of the \`ConstructorIOClient\`, you can pass a \`cioJsClient\` instead of an \`apiKey\` to request results from constructor's servers
 
 > Note: when we say \`cioJsClient\`, we are referring to an instance of the [constructorio-client-javascript](https://www.npmjs.com/package/@constructor-io/constructorio-client-javascript)`;
+export const cioJsClientOptionsDescription = `If you don't want to create an instance of the \`ConstructorIOClient\` but still want to customize some of the options, you can pass a \`cioJsClientOptions\` object. You can learn more about the possible values [here under the parameters section](https://constructor-io.github.io/constructorio-client-javascript/ConstructorIO.html).`;
```
esezen, line 114: I don't love this description but I wanted to keep it short. Please feel free to make or suggest edits

src/hooks/useCioAutocomplete.ts
```diff
@@ -37,7 +38,7 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
 
   const [query, setQuery] = useState('');
   const previousQuery = usePrevious(query);
-  const cioClient = useCioClient({ apiKey, cioJsClient } as CioClientConfig);
+  const cioClient = useCioClient({ apiKey, cioJsClient, cioJsClientOptions } as CioClientConfig);
```
mocca102, line 41: Since API key is a cioJsClientOption. What do you think about deprecating `apiKey` as a standalone parameter and make it passed as part of `cioJsClientOptions`? Or is it better to just leave it separate?

src/hooks/useCioClient.ts
```diff
@@ -6,13 +6,16 @@ import { CioClientConfig } from '../types';
...
-const useCioClient: UseCioClient = ({ apiKey, cioJsClient }) => {
+const useCioClient: UseCioClient = ({ apiKey, cioJsClient, cioJsClientOptions }) => {
   if (!apiKey && !cioJsClient) {
     // eslint-disable-next-line no-console
     console.error('Either apiKey or cioJsClient is required');
   }
 
-  return useMemo(() => cioJsClient || getCioClient(apiKey), [apiKey, cioJsClient]);
+  return useMemo(
+    () => cioJsClient || getCioClient(apiKey, cioJsClientOptions),
```
mocca102, line 16: Here we will just pass cioJsClientOptions which will include the apiKey

src/hooks/useCioAutocomplete.ts
esezen: I think it's better to leave it separate. I feel like most of the customers won't use `cioJsClientOptions` unless they are making a hybrid integration. I think `apiKey` makes it way easier to understand and get up an running with the library

mocca102: Yeah probably it makes more sense to library consumers that don't necessarily need to know that this is a config for the client

mocca102 (approved):
LGTM!

PR #112, opened 2023-12-11
[CSL-2867] Section configuration option updates

esezen:


src/types.ts
```diff
@@ -75,34 +76,52 @@ type SectionType = 'autocomplete' | 'recommendations' | 'custom';
...
   displaySearchTermHighlights?: boolean;
 };
 
-export interface AutocompleteSection extends SectionConfiguration {
+export interface AutocompleteSectionConfiguration extends SectionConfiguration {
   type?: 'autocomplete';
-  data: Item[];
+  indexSection: string;
+  // TODO: Add documentation about deprecation
+  identifier?: string;
```
esezen, line 89: Remove this

src/components/Autocomplete/SectionItemsList/SectionItemsList.tsx
```diff
@@ -15,12 +15,35 @@ type SectionItemsListProps = {
 
 // eslint-disable-next-line func-names
 const DefaultRenderSectionItemsList: RenderSectionItemsList = function ({ section }) {
-  const sectionName = section?.displayName || section?.identifier;
+  const { type, displayName } = section;
+  let sectionName = displayName;
```
esezen, line 19: Rename this to `sectionTitle`

src/components/Autocomplete/SectionItemsList/SectionItemsList.tsx
```diff
@@ -15,12 +15,35 @@ type SectionItemsListProps = {
...
+        break;
+    }
+  }
 
   if (!section?.data?.length) return null;
 
+  // Add the indexSection as a class to the section container to make sure it gets the styles
+  // Even if the section is a recommendation pod, if the results are "Products" or "Search Suggestions"
+  // ... they should be styled accordingly
+  const recommendationsSection = type === 'recommendations' ? section.indexSection : '';
```
esezen, line 43: Handle autocomplete here

esezen:
```
  const indexSection = type !== 'custom' ? kebab(section.indexSection) : '';
  ```

**Comment by mocca102** [comment]
This will need to be updated based on the introduced changes. It's using `section.identifier` which is deprecated in this PR.

https://github.com/Constructor-io/constructorio-ui-autocomplete/blob/b3305afa3d696abfc58578bfac874aaef70e0604/src/hooks/useCioAutocomplete.ts#L184

**Comment by mocca102** [comment]
I believe we need to add a type guard here. I believe this should work` if (!isCustomSection(section))`

https://github.com/Constructor-io/constructorio-ui-autocomplete/blob/b3305afa3d696abfc58578bfac874aaef70e0604/src/utils.ts#L196

Screenshot 2023-12-26 at 10 13 21 PM

File: src/types.ts [hunk 4]
```diff
@@ -75,35 +75,53 @@ type SectionType = 'autocomplete' | 'recommendations' | 'custom';
...
   // This property will only take effect when using the component and not the hook
   displaySearchTermHighlights?: boolean;
   ref?: React.RefObject<HTMLElement>;
 };
 
-export interface AutocompleteSection extends SectionConfiguration {
+export interface AutocompleteSectionConfiguration extends SectionConfiguration {
   type?: 'autocomplete';
-  data: Item[];
+  indexSection: string;
```
mocca102, line 87: Are we renaming this to indexSectionName for clarity?

src/types.ts
```diff
@@ -75,35 +75,53 @@ type SectionType = 'autocomplete' | 'recommendations' | 'custom';
...
+
+export interface AutocompleteSection extends AutocompleteSectionConfiguration {
+  data: Item[];
+}
+
+export interface RecommendationsSection extends RecommendationsSectionConfiguration {
+  data: Item[];
+}
+
+export interface CustomSection extends CustomSectionConfiguration {
```
mocca102, line 115: Having section vs sectionConfiguration make sense for the other sections that gets data added to it later. But for CustomSectionConfiguration `data` is already part of the configuration. What do you think about adding `data: Item[];` to `CustomSectionConfiguration` and remove `CustomSection` then replace occurrences of `CustomSection` -> `CustomSectionConfiguration`

src/types.ts
```diff
@@ -75,35 +75,53 @@ type SectionType = 'autocomplete' | 'recommendations' | 'custom';
...
   data: Item[];
 }
 
 export type Section = AutocompleteSection | RecommendationsSection | CustomSection;
 
-export type UserDefinedSection = CustomSection | SectionConfiguration;
+export type UserDefinedSection =
+  | AutocompleteSectionConfiguration
+  | RecommendationsSectionConfiguration
+  | CustomSection;
```
mocca102, line 124: here CustomSection -> CustomSectionConfiguration

src/utils.ts
```diff
@@ -53,7 +53,14 @@
     // uppercase the first character
     .replace(/^./, (str) => str.toUpperCase());
 
-export function isTrackingRequestSent(trackingRequestUrl) {
+export const toKebabCase = (str: string): string =>
+  str
+    .replace(/([A-Z]+)([A-Z][a-z])/g, '$1-$2')
```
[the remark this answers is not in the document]
mocca102: I think we will need to address this

src/typeGuards.ts
```diff
@@ -20,5 +20,5 @@ export function isInGroupSuggestion(item: Item): item is InGroupSuggestion {
 }
 
 export function isCustomSection(config: UserDefinedSection): config is CustomSection {
```
mocca102, line 22: Can we add other type guards as this one for `isAutocompleteSection` and `isRecommendationSection`?

src/types.ts
```diff
@@ -75,35 +75,53 @@ type SectionType = 'autocomplete' | 'recommendations' | 'custom';
...
+  identifier?: string;
 }
 
-export interface RecommendationsSection extends SectionConfiguration {
+export interface RecommendationsSectionConfiguration extends SectionConfiguration {
   type: 'recommendations';
-  data: Item[];
+  indexSection: string;
+  podId: string;
   itemIds?: string[];
```
mocca102, line 96: This is old but I don't believe it's accurate. We don't use itemIds as part of the section configuration. We attach the items from the data information [here](https://github.com/Constructor-io/constructorio-ui-autocomplete/blob/b3305afa3d696abfc58578bfac874aaef70e0604/src/utils.ts#L209) for `trackRecommendationView`

Do you think `itemIds?: string[];` has any value here or smth I might be missing?

src/types.ts
```diff
@@ -75,35 +75,53 @@ type SectionType = 'autocomplete' | 'recommendations' | 'custom';
...
-  section?: string;
   term?: string;
```
mocca102, line 97: Same for term

src/types.ts
```diff
@@ -75,35 +75,53 @@ type SectionType = 'autocomplete' | 'recommendations' | 'custom';
...
+  data: Item[];
+}
+
+export interface RecommendationsSection extends RecommendationsSectionConfiguration {
+  data: Item[];
+}
+
+export interface CustomSection extends CustomSectionConfiguration {
   data: Item[];
 }
```
mocca102, line 118: The above is becoming confusing, redundant and introduces unnecessary types that makes it hard to understand the different configs 😭 
Screenshot 2023-12-26 at 10 58 14 PM

What do you think of adding a utility type

- ` type WithData<T> = T & { data: Item[] };`
- `export type Section = WithData<AutocompleteSectionConfiguration> | WithData<RecommendationsSectionConfiguration> | WithData<CustomSectionConfiguration>;`

This will trim down types especially that they're not used out of this file

src/hooks/useSections.ts
```diff
@@ -24,14 +30,14 @@ export default function useSections(
     () =>
       activeSections?.filter(
         (config: UserDefinedSection) => config.type === 'autocomplete' || !config.type
```
mocca102, line 32: We implicitly consider a section without type to be an autocomplete section. 
[Here's another example](https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/112/files#diff-c72a6a39c2792562569f2b7e56003020f1eefbc541ba6221750d044635f0d1b1R32)

Which let me believe it's worth making it required or always default it to 'autocomplete' if it's not provided instead

src/types.ts
```diff
@@ -75,35 +75,53 @@ type SectionType = 'autocomplete' | 'recommendations' | 'custom';
...
   displayName?: string;
   numResults?: number;
   // This property will only take effect when using the component and not the hook
   displaySearchTermHighlights?: boolean;
   ref?: React.RefObject<HTMLElement>;
 };
 
-export interface AutocompleteSection extends SectionConfiguration {
+export interface AutocompleteSectionConfiguration extends SectionConfiguration {
   type?: 'autocomplete';
```
mocca102, line 86: What do you think about making type here required as well. For context
https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/112/files#r1436618676

src/hooks/useCioAutocomplete.ts
```diff
@@ -14,24 +14,48 @@ import {
...
+      }
+
+      if (!config.indexSection) {
+        return { ...config, indexSection: 'Products' };
+      }
+    }
+
+    if (type === 'autocomplete' || !type) {
+      if (config.identifier && !config.indexSection) {
+        return { ...config, indexSection: config.identifier };
```
mocca102, line 52: Maybe we can add `type: 'autocomplete'` here when `!type` instead of making it required

Context https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/112/files#r1436618676

src/types.ts
esezen: Are we?

src/types.ts
esezen: This makes sense. I was trying to keep the change non-breaking. Considering we don't have anyone using a custom section at the moment, I guess we can make a breaking change

src/typeGuards.ts
esezen: We can but why? If we are not using them, why are we adding them? For consumers of the library?

src/types.ts
esezen: I believe `term` and `itemIds` are parameters for the `getRecommendations` calls not for tracking

src/types.ts
esezen: I'm not against this change but the code you posted seems more confusing than what we have. It's shorter but it's more confusing. Let's talk about this today

src/hooks/useSections.ts
esezen: Where would we add that default?

src/types.ts
esezen: The type should be optional to make it easier for consumers to use the library. If nothing is passed, we assume autocomplete.  If anything, we can add a default on our end

src/hooks/useCioAutocomplete.ts
esezen: We can, I'm just not entirely sure how we would type it. Let's pair on it

src/typeGuards.ts
esezen: Add type guards

src/types.ts
esezen: Leave the names alone but add comments

mocca102 (approved):
LGTM!

PR #113, opened 2023-12-12
[NO-CSL] Update camelToStartCase function

esezen:
Before
image

After
image

I think we should also revisit this function later because it assumes the starting string is camelCase but that's not always the case


HHHindawy (approved):
Thanks for updating this!

PR #115, opened 2023-12-14
[CSL-3081] Support custom translations

HHHindawy:


src/stories/Autocomplete/Component/AdvancedParameters.stories.tsx
```diff
@@ -97,3 +98,21 @@ addComponentStoryDescription(
...
 );
+
+export const Translations = ComponentTemplate.bind({});
+Translations.args = {
+  apiKey,
+  onSubmit,
+  advancedParameters: {
+    numTermsWithGroupSuggestions: 1,
+    numGroupsSuggestedPerTerm: 2,
+    translations: {
```
mocca102, line 109: This looks just like passing key strings that it doesn't make sense to call it translations.

What's the difference between this and passing a param called "inTextValue" for example

I know we discussed this a little earlier but I'd imagine translations to accept different languages and handle the logic for rendering the matching language for me not just taking in one language

mocca102: I believe it should at least accept 

{
   "en": {"in": "inside"},
    "it": {"in": "..."}
}

HHHindawy: @mocca102 I guess that's what we agreed on earlier 😅 
@esezen correct me if I'm wrong; We agreed that we'd take in an object with direct translations of our translatable words (in our case just `in` for now), rather than taking in the whole translations JSON object and the current language (2 parameters).

There's no point of taking in translations for all languages since only one language will be active at any certain point of time. 

I'll bring up Enes's example once again.

```
const lang = "en" // Current language known by consumer at all times
const locales = import('locales.json');  //   Consumer's JSON file that has all translations
//   "en": {
//      "in": "inside"
//    },
//    "fr": {
//      "in": "..."
//    }

<CioAutocomplete
  language="en"
  translations={locales}
/>
```

So now rather than we get this entire json file and language consumers can pass us the snippet of this json for the current language and that's it:

```
const lang = "en" // Current language known by consumer at all times
const locales = import('locales.json');  //   Consumer's JSON file that has all translations
//   "en": {
//      "in": "inside"
//    },
//    "fr": {
//      "in": "..."
//    }

<CioAutocomplete
  translations={locales[lang]}
/>
```

> What's the difference between this and passing a param called "inTextValue" for example?

Well, since we only have a single word to be translated there's no much of a difference, but later on if we added more words then the object structure will come in handy!

esezen: @HHHindawy How did you explain my explanation better than I did? 

@mocca102 I thought that's what we agreed on but I'm open to discussing it further in one of the pairing sessions

src/hooks/useTranslate.ts
```diff
@@ -0,0 +1,17 @@
+import { Translations } from '../types';
+
+const useTranslate = (translations?: Translations) => {
```
esezen, line 3: Can we move this to utils and change the name format? `useXXX` implies this is a custom hook but it's not

src/constants.ts
```diff
@@ -229,3 +229,5 @@ To see this in action:
 export const termsWithImagesAndCountsDescription = `Pass boolean flags for \`displaySearchSuggestionImages\` and \`displaySearchSuggestionResultCounts\` fields to display images and counts for search suggestions. These fields need to be made displayable before they can be used. Please contact your Constructor Integration Engineer for details.`;
 
 export const debounceDescription = `Pass an integer to \`debounce\` to override the recommended, default delay employed for debouncing autocomplete network requests between keystrokes as your users type into the text input field. The default value is 250, which results in a debounce delay of 250 milliseconds.`;
+
+export const translationsDescription = `Pass a \`translations\` object to display translatable words in your preferred language.`;
```
esezen, line 233: Not sure where but can we document these translatable words somewhere? As a use I want to know all of the words I should be passing translations for

src/types.ts
```diff
@@ -124,3 +125,7 @@ export type HTMLPropsWithCioDataAttributes<T = any> = React.DetailedHTMLProps<
 > & {
   [key: `data-cnstrc-${string}`]: any;
 };
+
+export type Translations = {
+  in?: String;
```
esezen, line 130: Should this be lowercase string?

src/stories/Autocomplete/Component/AdvancedParameters.stories.tsx
```diff
@@ -97,3 +98,21 @@ addComponentStoryDescription(
...
+      in: 'inside',
```
esezen, line 110: Can we use a different language in our example so people understand the use case better?

esezen (changes requested):

esezen (approved):
LGTM! Great work 🚀

PR #117, opened 2023-12-19
Update storybook version to v7.6.6

mocca102:


src/stories/Autocomplete/Component/index.stories.tsx
```diff
@@ -1,5 +1,6 @@
+import type { Meta, StoryObj } from '@storybook/react';
 import ConstructorIOClient from '@constructor-io/constructorio-client-javascript';
-import { CioAutocomplete } from '../../../index';
+import { CioAutocomplete, CioAutocompleteProps } from '../../../index';
 import { argTypes } from '../argTypes';
```
esezen, line 4: Lint

src/types.ts
```diff
@@ -102,7 +102,7 @@ export interface CustomSection extends SectionConfiguration {
 
 export type Section = AutocompleteSection | RecommendationsSection | CustomSection;
 
-export type UserDefinedSection = CustomSection | SectionConfiguration;
+export type UserDefinedSection = Section;
```
esezen, line 105: Why are we changing types on a Storybook PR? I think this is fixed in another PR already so maybe we can skip it here?

esezen (changes requested):
This is looking good 🚀 I think we should revert the type changes, let me know if you disagree

esezen (approved):
LGTM!

PR #118, opened 2023-12-21
Add default value for autocomplete input

mocca102:


stanlp1 (approved):
This is looking good!

src/hooks/useCioAutocomplete.ts
```diff
@@ -34,9 +34,10 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
     zeroStateSections,
     autocompleteClassName = 'cio-autocomplete',
     advancedParameters,
+    defaultValue,
```
HHHindawy, line 37: I think it would be better if we named this prop: `defaultInput` or `defaultQuery`

esezen (approved):
LGTM!

PR #119, opened 2023-12-21
Explicitly export types path in package.json exports so it can be resolved

mocca102:
To reproduce the error that was caused by the absence of this

- npx create-next-app@latest
- cd your-app
- npm i @constructor-io/constructorio-ui-autocomplete
- Create a component and paste in the following snippet (in thread)
- Observe error in ticket in import statement for useCioAutocomplete

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,11 +11,13 @@
     },
     ".": {
       "import": "./lib/mjs/index.js",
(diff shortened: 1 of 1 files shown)


esezen (approved):
LGTM!

PR #121, opened 2023-12-26
Add AutocompleteSubmitEvent type predicates

mocca102:
onSubmit event takes a union type as a parameter. Requiring that you add type guards before accessing any of the event object properties. Since this is a very common event. This PR Adds those type predicates with usage examples


esezen (approved):
LGTM!

PR #124, opened 2024-01-05
[NO-CSL] Update fully featured example styles

esezen:


stanlp1 (approved):
This change lgtm.

I noticed that after the section name changes, we now add classNames with spaces in them. Is this intentional/does it matter?

Screenshot 2024-01-04 at 11 24 07 PM

PR #126, opened 2024-01-18
[CSL-3168] Surface the Display Name of Recommendations Pod by Default

Mudaafi:
## Summary
Recommendations Pod have a "display_name" attribute that we would want to render by default if available. If the user has specified a `displayName` configuration for the pod via the library, we should use that instead.
The css classnames should still reflect pod-id to avoid weird looking classnames such as `Best Sellers` rather than `best-sellers`.

## Changes Made
- `useFetchRecommendationsPod` now returns an additional object `podsData` which contains both the `podId` and `displayName` returned by the response.
- Consequently, `utils/getActiveSectionsWithData` now takes in the `podsData` as a new optional parameter.
- In `getSectionProps` in `useCioAutocomplete`, `sectionTitle` is used setting classnames and should reflect `podId` rather than `displayName`
- In `SectionItemList`, `sectionTitle` here is used to render to the end user and should reflect the `displayName` for Recs Pods, if available.

### DIFF
--- a/src/hooks/useCioAutocomplete.ts
+++ b/src/hooks/useCioAutocomplete.ts
@@ -212,8 +212,8 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
       'data-testid': 'cio-form',
     }),
     getSectionProps: (section: Section) => {
-      const { type, displayName } = section;
-      let sectionTitle = displayName;
+      const { type } = section;
+      let sectionTitle: string;
 
       // Add the indexSectionName as a class to the section container to make sure it gets the styles
       // Even if the section is a recommendation pod, if the results are "Products" or "Search Suggestions"
@@ -221,21 +221,19 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
       const indexSectionName =
         type !== 'custom' && section.indexSectionName ? toKebabCase(section.indexSectionName) : '';
 
-      if (!sectionTitle) {
-        switch (type) {
-          case 'recommendations':
-            sectionTitle = section.podId;
-            break;
-          case 'autocomplete':
-            sectionTitle = section.indexSectionName;
-            break;
-          case 'custom':
-            sectionTitle = section.displayName;
-            break;
-          default:
-            sectionTitle = section.indexSectionName;
-            break;
-        }
+      switch (type) {
+        case 'recommendations':
+          sectionTitle = section.podId;
+          break;
+        case 'autocomplete':
+          sectionTitle = section.displayName || section.indexSectionName;
+          break;
+        case 'custom':
+          sectionTitle = section.displayName;
+          break;
+        default:
+          sectionTitle = section.displayName || section.indexSectionName;
+          break;
       }
 
       const attributes: HTMLPropsWithCioDataAttributes = {
--- a/src/hooks/useFetchRecommendationPod.ts
+++ b/src/hooks/useFetchRecommendationPod.ts
@@ -1,13 +1,14 @@
 import { useEffect, useState } from 'react';
 import ConstructorIOClient from '@constructor-io/constructorio-client-javascript';
 import { Nullable } from '@constructor-io/constructorio-client-javascript/lib/types';
-import { Item, SectionsData, RecommendationsSectionConfiguration } from '../types';
+import { Item, SectionsData, RecommendationsSectionConfiguration, PodData } from '../types';
 
 const useFetchRecommendationPod = (
   cioClient: Nullable<ConstructorIOClient>,
   recommendationPods: RecommendationsSectionConfiguration[]
 ) => {
-  const [recommendationResults, setRecommendationResults] = useState<SectionsData>({});
+  const [recommendationsResults, setRecommendationsResults] = useState<SectionsData>({});
+  const [podsData, setPodsData] = useState<Record<string, PodData>>({});
 
   useEffect(() => {
     if (!cioClient || !Array.isArray(recommendationPods) || recommendationPods.length === 0) return;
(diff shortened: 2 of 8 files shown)


esezen (commented):
This is looking good. The only thing I'm wondering is if there is a way to make `podsData` part of the other objects instead of passing it around individually. I couldn't find a good solution though, so I think what we have here is good. Just wanted to mention it to see if there are any other ideas. Not marking it as approved to make sure we get a review from @mocca102 before we merge but this LGTM!

src/hooks/useCioAutocomplete.ts
```diff
@@ -212,30 +212,28 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
       'data-testid': 'cio-form',
     }),
     getSectionProps: (section: Section) => {
-      const { type, displayName } = section;
-      let sectionTitle = displayName;
+      const { type } = section;
+      let sectionTitle: string;
```
mocca102, line 216: Can we rename this to sectionClassName instead? or anything that says this differs from the section's actual title? The name is confusing and coming back to this in the future we would forget that this is only for the class name

src/hooks/useFetchRecommendationPod.ts
```diff
@@ -20,22 +21,24 @@ const useFetchRecommendationPod = (
...
+      const recommendationsPodsData = {};
       responses.forEach(({ response }, index) => {
         const { pod, results } = response;
         if (pod?.id) {
-          recommendationPodResults[pod.id] = results?.map((item: Item) => ({
+          recommendationsPodResults[pod.id] = results?.map((item: Item) => ({
             ...item,
             id: item?.data?.id,
             section: recommendationPods[index]?.indexSectionName,
             podId: pod.id,
```
mocca102, line 33: We already did it once and added the podId. Can we also add sectionDisplayName here as well instead of having separate podData?

src/hooks/useCioAutocomplete.ts
Mudaafi: I'm in total agreement. @esezen I think you worked on this previously, but would `sectionClassName` work for you as well?

src/utils.ts
```diff
@@ -156,17 +156,20 @@ export const getCioClient = (apiKey?: string, cioJsClientOptions?: ConstructorCl
 export const getActiveSectionsWithData = (
   activeSections: UserDefinedSection[],
```
mocca102, line 157: This is where we have the sections configs and I believe it's the best place we can add the displayName for the recommendations active sections
Screenshot 2024-01-24 at 11 27 15 PM

src/hooks/useSections.ts
```diff
@@ -51,15 +51,17 @@ export default function useSections(
...
+  const { recommendationsResults, podsData } = useFetchRecommendationPod(
+    cioClient,
+    recommendationsSections
+  );
   // Add to active sections the results data and refs when autocomplete results or recommendation results fetched
   useEffect(() => {
     const sectionsResults = { ...autocompleteResults, ...recommendationsResults };
     setActiveSectionsWithData(
-      getActiveSectionsWithData(activeSections, sectionsResults, sectionsRefs)
+      getActiveSectionsWithData(activeSections, sectionsResults, sectionsRefs, podsData)
```
mocca102, line 62: Here is the place where we can get the data from recommendationsResults where the display name on the items and add it the appropriate active section instead of returning podsData then passing it

mocca102 (changes requested):
I left some comments. Let me know what do you think. It's just about how we pass the data around

src/hooks/useFetchRecommendationPod.ts
Mudaafi: We could but I don't particularly like it mainly because the sectionDisplayName would exist per item rather than at the top level which could get messy. From what I understand, `podId` needs to exist per item for tracking-related things but `displayName` doesn't so I would rather not put it there.

An alternative implementation, and this might attend to Enes' idea, we could modify the `recommendationsPodResults` to be an object rather than an array. That way we can return one object rather than two but this will be change the expected interface.

mocca102: I agree that modifying `recommendationsPodResults` would be the ideal thing here so I thought adding sectionDisplayName would be less intrusive and wouldn't require passing extra parameters.

I'm ok if you want to keep it the way you implemented though

src/hooks/useCioAutocomplete.ts
Mudaafi: We'll track this on PR #128 instead

src/hooks/useFetchRecommendationPod.ts
Mudaafi: We'll opt of merging it before calling `setActiveSectionsWithData`

mocca102 (approved):
Thanks for making the changes. LGTM!

src/utils.ts
```diff
@@ -182,8 +182,8 @@ export const getActiveSectionsWithData = (
 
     if (Array.isArray(sectionData)) {
       const section = {
-        ...sectionConfig,
```
esezen, line 185: @Mudaafi let me know if I'm thinking about this incorrectly but I think order of the spread (...) vs `data` is important here. For custom sections, we want new `sectionData` to override the one coming from the config.
Screenshot 2024-01-27 at 1 34 45 PM
Screenshot 2024-01-27 at 1 33 49 PM

esezen (changes requested):

src/utils.ts
mocca102: Should sectionConfig have data for other sections than custom? I think it shouldn't 🤔

esezen: Correct, only for `custom`

mocca102: So why would we worry if sectionConfig overrides sectionData then?

esezen: Because we don't want that for custom sections and this code runs for custom sections? The screenshot above is from a custom section. Am I understanding this wrong?

mocca102: What I mean is it's ok for custom sections if it overrides the data since we're not handling it anyway

esezen: On line 173 we are modifying the data being passed to us then here we are overriding it back

mocca102: Aha got it

Mudaafi: Just so I'm on the same page, are we saying the order of priority is:
1. Dashboard/Catalog set configs
2. Developer-set top-level-component configs

That didn't feel intuitive to me :thinking: but maybe it's because I'm thinking of it as an "override" rather than a "fallback"

Mudaafi: Okay nvm scratch that, I get the issue with `data` getting overriden

esezen (approved):
LGTM!

PR #128, opened 2024-01-24
[CSL-3088] - fix issue with conflicting styles

martinrajdl:
What this change include? 
- removes sentence cased classnames from section list items & moves to `data-cnstrc-display-name` 
- prefixes kebab cased classname like so `cio-section-kebab-case-item-section` 
- keeps main classname 
Before: 
Screenshot 2024-01-24 at 17 16 33
After: 
Screenshot 2024-01-24 at 17 08 54
- removes unnecessary classname from section item 
Screenshot 2024-01-24 at 18 16 28

- task also includes criteria to "Make sectionName camelcase" but it already is 🤔

### DIFF
--- a/src/stories/Autocomplete/Hook/index.tsx
+++ b/src/stories/Autocomplete/Hook/index.tsx
@@ -3,7 +3,7 @@ import React from 'react';
 import useCioAutocomplete from '../../../hooks/useCioAutocomplete';
 import { isRecommendationsSection } from '../../../typeGuards';
 import { Item } from '../../../types';
-import { getStoryParams, toKebabCase } from '../../../utils';
+import { camelToStartCase, getStoryParams, toKebabCase } from '../../../utils';
 
 export function HooksTemplate(args) {
   const {
@@ -119,32 +119,38 @@ export function HooksTemplate(args) {
             if (!section?.data?.length) {
               return null;
             }
+
             const { type, displayName } = section;
-            let sectionName = section.displayName;
+            let sectionTitle: string;
 
             switch (type) {
               case 'recommendations':
-                sectionName = section.podId;
+                sectionTitle = section.podId;
                 break;
               case 'custom':
-                sectionName = toKebabCase(displayName);
+                sectionTitle = displayName;
                 break;
               case 'autocomplete':
-                sectionName = section.indexSectionName;
+                sectionTitle = section.indexSectionName;
                 break;
               default:
-                sectionName = section.indexSectionName;
+                sectionTitle = section.indexSectionName;
                 break;
             }
 
-            const recommendationsSection = isRecommendationsSection(section)
-              ? section.indexSectionName
-              : '';
+            if (displayName) {
+              sectionTitle = displayName;
+            }
+
+            const sectionClassNames = toKebabCase(sectionTitle);
+            if (isRecommendationsSection(section)) {
+              sectionClassNames.concat(` ${toKebabCase(section.indexSectionName)}`);
+            }
 
             return (
-              <div key={sectionName} className={`${sectionName} ${recommendationsSection}`}>
+              <div key={sectionTitle} className={sectionClassNames}>
                 <div {...getSectionProps(section)}>
-                  <h5 className='cio-section-name'>{sectionName}</h5>
+                  <h5 className='cio-section-name'>{camelToStartCase(sectionTitle)}</h5>
                   <div className='cio-section-items'>
                     {section?.data?.map((item) => renderItem(item))}
                   </div>
--- a/src/stories/Autocomplete/Hook/index.tsx
+++ b/src/stories/Autocomplete/Hook/index.tsx
@@ -142,9 +142,9 @@ export function HooksTemplate(args) {
               sectionTitle = displayName;
             }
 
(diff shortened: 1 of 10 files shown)


src/stories/tests/HooksTests.stories.tsx
```diff
@@ -219,6 +219,7 @@ TypeSearchTermRenderSectionsCustomOrder.args = {
     },
   ],
 };
+// TODO: duplicate tests? these don't seem to run on npm run test-storybook
 TypeSearchTermRenderSectionsCustomOrder.play = async ({ canvasElement }) => {
```
Mudaafi, line 223: These do run but they're aren't using the Components we export. These classes are found in the `/src/stories/Autocomplete/Hook/index.tsx#HooksTemplate`. We should probably update them there to be consistent in terms of our docs.

Mudaafi (changes requested):
This is looking good, thanks for working on this! 

Since we're surfacing `sectionTitle` as `data-cnstrc-display-name`, lets add some tests to check for them as well. We should probably rename `sectionTitle` to `sectionClassName` as well to add clarity.

src/stories/Autocomplete/Hook/index.tsx
```diff
@@ -119,32 +119,38 @@ export function HooksTemplate(args) {
...
+                sectionTitle = displayName;
                 break;
               case 'autocomplete':
-                sectionName = section.indexSectionName;
+                sectionTitle = section.indexSectionName;
                 break;
               default:
-                sectionName = section.indexSectionName;
+                sectionTitle = section.indexSectionName;
                 break;
```
Mudaafi, line 138: This is (type) broken currently but we still need it. Sections can be one of type `autocomplete` | `recommendations` | `custom`, but the type on `autocomplete` is optional meaning it can be `undefined`.

There's two ways to fix this:
 1. Define an additional type that mimics `AutocompleteConfiguration` but with no type
 2. Empty the default clause here. Add an if-conditional above to check if `type===undefined`
 3. Assign the type during `convertLegacyParametersAndAddDefaults` in `useCioAutocomplete` (preferred)

I'll pull up a separate PR for this.

mocca102: which part you mean "This is (type) broken currently"? @Mudaafi

src/stories/Autocomplete/Hook/index.tsx
```diff
@@ -119,32 +119,41 @@ export function HooksTemplate(args) {
             if (!section?.data?.length) {
               return null;
             }
+
             const { type, displayName } = section;
-            let sectionName = section.displayName;
+            const sectionListingType = isCustomSection(section)
```
mocca102, line 124: Is there a way we can get this from getSectionProps since this same logic is being duplicated 
https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/128/files#diff-7f1c987b478566cc9cf82a2afc2ee7a7a150ec6047a7801e0ac01c410f7ce453R218

src/stories/Autocomplete/Hook/index.tsx
```diff
@@ -119,32 +119,41 @@ export function HooksTemplate(args) {
...
+              ? 'custom'
+              : toKebabCase(section.indexSectionName || section.data[0]?.section || 'Products');
 
+            let sectionTitle: string;
```
mocca102, line 128: Same for sectionTitle. Is there a way we can get this from getSectionProps instead?

mocca102 (commented):
Styles not affected, Logic looks solid, just left two comments

src/stories/Autocomplete/Hook/index.tsx
Mudaafi: @mocca102 Might be my configuration, but since we've explicitly defined the type `Section` to be one of three things and the switch statement accounts for all three, the final default type is `never`. TS then throws the error `Property 'indexSectionName' does not exist on type 'never'`

src/stories/Autocomplete/Hook/index.tsx
Mudaafi: We could. The reason why its included here is because we've rendered a separate `div` with the relevant classes in [Line 154](https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/128/files#diff-0c9a1e20dad0ff16de33515c80eb5efdfda99bebefc29a58edf246f5d017c5dfR154) so we kind of have to duplicate the logic to get the required values here.

Took a look at our history and it might be something we've included prior to creating `getSectionProps` so we might be good to omit this entirely from the story. @esezen would love to get your thoughts here since you [touched it previously](https://github.com/Constructor-io/constructorio-ui-autocomplete/commit/3ad2d46789d886a966c15a80c774d6c57d74fd91#diff-0c9a1e20dad0ff16de33515c80eb5efdfda99bebefc29a58edf246f5d017c5dfR145).

Mudaafi: In terms of passing this variable via `getSectionProps` itself, I thought `getSectionProps` is meant to contain only the attributes we're going to render into the HTML? I would be against adding something in specifically for our Storybook.

esezen (approved):
LGTM!

Mudaafi (approved):

PR #129, opened 2024-02-09
[CSL-3053] Zero state recommendations a/b support

stanlp1:


src/hooks/useSections.ts
```diff
@@ -56,6 +56,14 @@ export default function useSections(
     recommendationsSections
   );
 
+  // Delete active sections if feature toggle, zeroStateActiveSessions is present
+  if (zeroStateActiveSections && podsData && Object.keys(podsData).length) {
+    const features = getSearchSuggestionFeatures(Object.values(podsData)[0]);
+    if (features.featureDisplayZeroStateRecommendations) {
+      activeSections.length = 0;
```
stanlp1, line 63: Not too sure I like what I'm doing here.

kwekke: there could be side effects when setting `activeSections.length = 0` since it modifies the original array. you might want to have immutability for cleaner and predictable code. 

consider this scenario
const sections = [a,b,c];
console.log(sections) // prints [a,b,c] 
const {...} = useSections(query, cioClient, sections);
// it runs `activeSections.length = 0`
console.log(sections) /// prints [] 


off the top of my head, you could instead do 
const [activeSections, setActiveSections] = useState(zeroStateActiveSections ? zeroStateSections : sections);
useEffect(() => {
  // Delete active sections if feature toggle, zeroStateActiveSessions is present
  if (...) {
    setActiveSections([]);
  }
}, [zeroStateActiveSections, podsData, getSearchSuggestionFeatures])

stanlp1: Thanks, that is definitely conforms to the react pattern better.

src/hooks/useSections.ts
```diff
@@ -9,7 +9,7 @@ import {
   UserDefinedSection,
   Section,
 } from '../types';
-import { getActiveSectionsWithData } from '../utils';
+import { getActiveSectionsWithData, getSearchSuggestionFeatures } from '../utils';
```
esezen, line 12: Should we consider renaming `getSearchSuggestionFeatures`? It also applies to rex now

src/hooks/useSections.ts
```diff
@@ -56,6 +59,21 @@ export default function useSections(
     recommendationsSections
   );
 
+  // Remove sections if necessary
+  useEffect(() => {
+    const features = getSearchSuggestionFeatures(Object.values(podsData || {})?.[0]?.request);
+
+    if (zeroStateActiveSections) {
+      if (features.featureDisplayZeroStateRecommendations) {
+        setActiveSections([]);
```
esezen, line 68: Do we need any of the elses? Wouldn't they come from the initial assignment 🤔

esezen (commented):
This is looking good. I left a few comments, let me know what you think

src/hooks/useSections.ts
stanlp1: I think we will need the else to setActiveSections(sections), but we may not need the one for setActiveSections(zeroStateSections), if we don't expect the a/b state to change in each request.

Correct me if I am wrong but 

```
  const [activeSections, setActiveSections] = useState<UserDefinedSection[]>(
    zeroStateActiveSections ? zeroStateSections : sections
  );
```

Only runs once, so if we don't have the else for setActiveSections(sections), the autocomplete will always be empty.

src/hooks/useSections.ts
stanlp1: Thoughts on name? Thinking of just "getFeatures"

esezen: I think that works 👍

src/hooks/useSections.ts
esezen: I think you are right. Thanks for the explanation

esezen (approved):
LGTM! I think the tests are failing because we have the feature toggle on. Can you turn the toggle of and run tests on the PR before you merge?

PR #139, opened 2024-05-02
Save search submit to local storage

mocca102:


src/beaconUtils.js
```diff
@@ -0,0 +1,181 @@
+export const CONSTANTS = {
```
mocca102, line 1: Copied from the beacon code

src/utils.ts
```diff
@@ -4,6 +4,7 @@ import {
   Nullable,
   ConstructorClientOptions,
 } from '@constructor-io/constructorio-client-javascript/lib/types';
+import { storageSetItem, storeRecentSearch, storeRecentAction, CONSTANTS } from './beaconUtils';
```
esezen, line 7: Lint

src/utils.ts
```diff
@@ -256,3 +257,20 @@ export const translate = (word: string, translations?: Translations) => {
 
   return localTranslations[word] || word;
 };
+export const addTrackSearchSubmit = (cioClient, term, autocompleteData = {}) => {
```
esezen, line 260: Naming thing we talked about

esezen (changes requested):

esezen (approved):
LGTM! Let's get one more review

PR #141, opened 2024-05-10
[CI-3436] Autocomplete UI - Add an option to call zero state requests on focus

lordvkrum:
* Add advancedParameters fetchZeroStateOnFocus to override the zero state fetching behavior from initial render to input focus

### DIFF
--- a/src/hooks/useFetchRecommendationPod.ts
+++ b/src/hooks/useFetchRecommendationPod.ts
@@ -5,54 +5,58 @@ import { Item, SectionsData, RecommendationsSectionConfiguration, PodData } from
 
 const useFetchRecommendationPod = (
   cioClient: Nullable<ConstructorIOClient>,
-  recommendationPods: RecommendationsSectionConfiguration[]
+  recommendationPods: RecommendationsSectionConfiguration[],
+  fetchZeroStateOnFocus: boolean = false
 ) => {
   const [recommendationsResults, setRecommendationsResults] = useState<SectionsData>({});
   const [podsData, setPodsData] = useState<Record<string, PodData>>({});
 
-  useEffect(() => {
+  const fetchRecommendationResults = async () => {
     if (!cioClient || !Array.isArray(recommendationPods) || recommendationPods.length === 0) return;
-    const fetchRecommendationResults = async () => {
-      const responses = await Promise.all(
-        recommendationPods.map(({ podId, indexSectionName, ...parameters }) =>
-          cioClient.recommendations.getRecommendations(podId, {
-            ...parameters,
-            section: indexSectionName,
-          })
-        )
-      );
-      const recommendationsPodResults = {};
-      const recommendationsPodsData = {};
-      responses.forEach(({ response, request }, index) => {
-        const { pod, results } = response;
-        if (pod?.id) {
-          recommendationsPodResults[pod.id] = results?.map((item: Item) => ({
-            ...item,
-            id: item?.data?.id,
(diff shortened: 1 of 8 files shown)


src/hooks/useFetchRecommendationPod.ts
```diff
@@ -5,54 +5,58 @@ import { Item, SectionsData, RecommendationsSectionConfiguration, PodData } from
...
   cioClient: Nullable<ConstructorIOClient>,
-  recommendationPods: RecommendationsSectionConfiguration[]
+  recommendationPods: RecommendationsSectionConfiguration[],
+  fetchZeroStateOnFocus: boolean = false
 ) => {
   const [recommendationsResults, setRecommendationsResults] = useState<SectionsData>({});
   const [podsData, setPodsData] = useState<Record<string, PodData>>({});
 
-  useEffect(() => {
+  const fetchRecommendationResults = async () => {
```
lordvkrum, line 14: moving function outside useEffect so it can be exported

src/types.ts
```diff
@@ -26,6 +26,7 @@ export interface AdvancedParametersBase {
   displaySearchSuggestionResultCounts?: boolean;
   debounce?: number;
   translations?: Translations;
+  fetchZeroStateOnFocus?: boolean;
```
lordvkrum, line 29: would having two props instead of a single one that replaces load with focus events be better?
ex:
fetchZeroStateOnFocus: default false
fetchZeroStateOnInit: default true

src/hooks/useFetchRecommendationPod.ts
mocca102: I like this! What do you think about moving it outside of useFetchRecommendationPod into utils file. So it can be imported in useCioAutocomplete without having to keep exposing it up across different hooks?

src/types.ts
mocca102: Lets discuss this with the team

mocca102 (commented):
This is looking good for the most part. Just left two comments, let me know what you think

src/types.ts
lordvkrum: After team discussion, we will leave the implementation as is with a single option, having two options may lead to scenarios where both are TRUE and that will cause duplicated tracking due to the wrong configuration

src/hooks/useFetchRecommendationPod.ts
mocca102: @lordvkrum I checked your POC branch, and I though of not calling fetchRecommendationResults inside the onFocus callback because of missing dependencies. And depend on the useEffect here to run an effect when the input is focused but there was still complexity with the dependencies. Because of that it's better that we move forward with your current implementation for now.

src/hooks/useRecommendationsObserver.ts
```diff
@@ -63,6 +66,13 @@ function useRecommendationsObserver(
       });
     };
     // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [menuIsOpen, sections]);
```
mocca102, line 69: There was a race condition where `menuIsOpen=true` but the recommendations element was not rendered and the function`trackRecommendationView` was not being called for the first focus but gets fired for subsequent focuses

Adding sections as dependency fixed that but caused multiple `trackRecommendationView` hence the addition if `viewedRecommendations` to keep track of when the recommendation has been viewed as long as the menu is open. Once the menu is closed it restrarts

esezen (approved):
LGTM!

PR #145, opened 2024-05-31
[CI-3476] Removes display none for product section breakpoint

stanlp1:


mocca102 (commented):
Do we need to handle mobile styles here (possibly stack them horizontally) or do this in a separate PR?

mocca102 (approved):
LGTM!

PR #154, opened 2024-07-01
[CI-3614] Don't send autocomplete requests with queries longer than 200 characters long

esezen:


src/hooks/useDebouncedFetchSections.ts
```diff
@@ -93,7 +93,11 @@ const useDebouncedFetchSection = (
 
   useEffect(() => {
     (async () => {
-      if (debouncedSearchTerm.trim()) {
+      // Do not send a request if the encoded term is longer than 199 characters
+      if (
+        debouncedSearchTerm.trim() &&
+        encodeURIComponent(debouncedSearchTerm.trim()).length < 200
+      ) {
```
jjl014, line 100: Should we save `debouncedSearchTerm.trim()` to a variable here?

jjl014 (approved):
Looks good from what I can tell! 😄 

Just left one minor comment.

src/hooks/useDebouncedFetchSections.ts
esezen: Optimization, I like it 🧠

jjl014 (approved):
LGTM! 🚀

PR #158, opened 2024-07-24
[NOCI] Fix infinite sections updates

HHHindawy:


jjl014 (commented):
This seems to be working great from what I can tell, but let's get another review from the team that's more familiar with the repo.

I was able to reproduce the issue (screenshot attached), but I wasn't sure how you tested the changes. Here are the steps that I took to test your changes:
* Pulled down the PR
* Run `npm run compile`
* Copy the generated `lib` directory and replaced the existing `lib` folder within the test project’s `@constructor-io/constructorio-ui-autocomplete` directory in `node_modules`.

Screenshot 2024-07-25 at 1 15 58 PM

mocca102 (changes requested):
I believe the issue persists for object parameters other than `sections` like `zeroStateSections`

esezen (approved):
LGTM!

PR #165, opened 2024-09-04
Refactor Suggestions components for clarity

mocca102:
This introduces no logic changes. It's just organizing code into Components for clarity.

Before

 - Everything was inside `SectionItem` 

After

 - Suggestion
    - SearchSuggestion
    - ProductSuggestion
    - CustomSuggestion  
    - InGroupSearchSuggestion

Added `useCioAutocompleteContext` to make it clear what context is being used

### DIFF
--- a/src/components/Autocomplete/SectionItem/SectionItemText.tsx
+++ b/src/components/Autocomplete/SectionItem/SectionItemText.tsx
@@ -2,18 +2,16 @@ import React from 'react';
 import { Item } from '../../../types';
 import { isInGroupSuggestion } from '../../../typeGuards';
 import { escapeRegExp } from '../../../utils';
+import useCioAutocompleteContext from '../../../hooks/useCioAutocompleteContext';
 
 export interface SectionItemTextProps {
   item: Item;
-  query: string;
   highlightSearchTerm?: boolean;
 }
 
-export default function SectionItemText({
-  item,
-  query,
-  highlightSearchTerm,
-}: SectionItemTextProps) {
+export default function SectionItemText({ item, highlightSearchTerm }: SectionItemTextProps) {
+  const { query } = useCioAutocompleteContext();
+
   const itemText = isInGroupSuggestion(item) ? item.groupName : item.value;
 
   if (highlightSearchTerm) {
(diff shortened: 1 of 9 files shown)


src/components/Autocomplete/SectionItemsList/SectionItemsList.tsx
```diff
@@ -14,36 +14,34 @@ type SectionItemsListProps = {
...
+    case 'custom':
+      return displayName;
+    default:
+      return section.indexSectionName;
   }
+};
+
+// eslint-disable-next-line func-names
+const DefaultRenderSectionItemsList: RenderSectionItemsList = function ({ section }) {
+  const { getSectionProps } = useContext(CioAutocompleteContext);
```
esezen, line 34: Since you already created a new hook, should we use it here? `useCioAutocompleteContext`

src/components/Autocomplete/SectionItemsList/SectionItemsList.tsx
```diff
@@ -14,36 +14,34 @@ type SectionItemsListProps = {
...
+  const { displayName } = section;
+  const sectionTitle = camelToStartCase(displayName || getSectionTitle(section));
```
esezen, line 36: I think we don't want to run `displayName` through `camelToStartCase.` If the user set it to be camelcase, we should leave it as is

src/components/Autocomplete/SectionItem/SectionItemText.tsx
```diff
@@ -2,18 +2,16 @@ import React from 'react';
...
   highlightSearchTerm?: boolean;
 }
 
-export default function SectionItemText({
-  item,
-  query,
-  highlightSearchTerm,
-}: SectionItemTextProps) {
+export default function SectionItemText({ item, highlightSearchTerm }: SectionItemTextProps) {
+  const { query } = useCioAutocompleteContext();
```
esezen, line 13: Q: Any reason why we don't move `displaySearchTermHighlights` here and stop passing it around?

esezen (commented):
This is looking great! Thanks for working on this, makes the code way more readable. I left small comments here and there

PR #166, opened 2024-09-11
Product suggestions are links by default

mocca102:
No breaking changes

Markup Before
Screenshot 2024-09-12 at 12 35 11 AM

Markup after
Screenshot 2024-09-12 at 12 36 06 AM

### DIFF
--- a/src/components/Autocomplete/SectionItem/SectionItem.tsx
+++ b/src/components/Autocomplete/SectionItem/SectionItem.tsx
@@ -27,7 +27,7 @@ export default function SectionItem(props: SectionItemProps) {
 
   if (isProduct(item)) {
     defaultChildren = (
-      <>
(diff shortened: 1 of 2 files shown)


src/components/Autocomplete/SectionItem/SectionItem.tsx
```diff
@@ -27,7 +27,7 @@ export default function SectionItem(props: SectionItemProps) {
 
   if (isProduct(item)) {
     defaultChildren = (
-      <>
+      <a href={item.data?.url}>
```
lordvkrum, line 30: small edge case, should we validate if there is no URL (not to render A element in that case)? clicking A with empty href will refresh the page

lordvkrum (approved):

PR #167, opened 2024-09-12
Add cnstrc data attributes

mocca102:


jjl014 (approved):
Looks good from what I can tell! I'm able to see the data attributes come through. 👍

I don't know enough about types to comment on those changes, but the changes seem to make sense

PR #172, opened 2024-09-27
[CI-3882] Do not fill input on Search Suggestion selection

ZSnake:
Remove setQuery from useDownShift to not re-set autocomplete value

### DIFF
--- a/src/hooks/useCioAutocomplete.ts
+++ b/src/hooks/useCioAutocomplete.ts
@@ -124,7 +124,7 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
     highlightedIndex,
     getInputProps,
     getItemProps: getItemPropsDownShift,
-  } = useDownShift({ setQuery, items, onSubmit, cioClient, previousQuery });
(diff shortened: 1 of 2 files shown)


Mudaafi (commented):
this lgtm. Just one note we can discuss tmr doing standup possibly:
1. Is this behavior always undesired? I can imagine some customers possibly liking it the way it is currently, specifically those utilizing a SPA.

Unrelated note (for other reviewers): The product card doesn't redirect here because it's branched from v1.23.10. The wrapper anchor tag was added in v1.23.11

mocca102:
> this lgtm. Just one note we can discuss tmr doing standup possibly:
> 
> 1. Is this behavior always undesired? I can imagine some customers possibly liking it the way it is currently, specifically those utilizing a SPA.
> 
> Unrelated note (for other reviewers): The product card doesn't redirect here because it's branched from v1.23.10. The wrapper anchor tag was added in v1.23.11

Thank you for raising the behavior for current library users I checked and I found some of our customers do utilize the current behavior so I do not think we can merge this

mocca102:
closed because it can break current customer implementation

PR #175, opened 2024-10-02
Fix zero state not being rendered when there is default input

mocca102:
Zero state did not work if `defaultInput` was passed

### DIFF
--- a/src/hooks/useFetchRecommendationPod.ts
+++ b/src/hooks/useFetchRecommendationPod.ts
@@ -1,7 +1,7 @@
 import { useEffect, useState } from 'react';
 import ConstructorIOClient from '@constructor-io/constructorio-client-javascript';
 import { Nullable } from '@constructor-io/constructorio-client-javascript/lib/types';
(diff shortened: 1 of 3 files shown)


src/hooks/useFetchRecommendationPod.ts
```diff
@@ -54,7 +54,7 @@ const useFetchRecommendationPod = (
     if (fetchZeroStateOnFocus) return;
     fetchRecommendationResults();
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [cioClient]);
+  }, [cioClient, recommendationPods]);
```
stanlp1, line 57: This is the main fix right? I got confused because of the other changes, wasn't sure how they were fixing the issue and missed this.

Correct me if I am wrong, but the change sin useSections don't change any behavior.

stanlp1 (approved):
LGTM.

src/hooks/useFetchRecommendationPod.ts
mocca102: You are correct I'm removing the unnecessary code changes

PR #176, opened 2024-10-10
Add getSearchResultsUrl util

mocca102:
Props
 - getSearchResultsUrl
  - It returns a string that when a user clicks on a search suggestion it redirects to the search results page for the selected term


stanlp1 (approved):
LGTM!

PR #180, opened 2024-11-04
[CI-3954] Allow to dynamically change filters on zero state

ZSnake:
- Remove check for items length on re-fetching recommendations results
- Add Zero State story and example

### DIFF
--- a/src/hooks/useCioAutocomplete.ts
+++ b/src/hooks/useCioAutocomplete.ts
@@ -190,7 +190,7 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
           openMenu();
         }
         try {
-          if (advancedParameters?.fetchZeroStateOnFocus && items?.length === 0) {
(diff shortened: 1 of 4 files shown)


src/hooks/useCioAutocomplete.ts
```diff
@@ -189,7 +189,7 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
           openMenu();
         }
         try {
-          if (advancedParameters?.fetchZeroStateOnFocus && items?.length === 0) {
+          if (advancedParameters?.fetchZeroStateOnFocus) {
```
mocca102, line 192: If you revert this change and test the storybook example you added everything will work the same. The recommendation pod get fetched when the filters change already I'm not even sure there's a problem to fix here 🤔 @ZSnake

mocca102 (approved):
LGTM!

PR #184, opened 2024-11-12
Add support for onIsOpenChange callback

mocca102:
- onIsOpenChange

### DIFF
--- a/src/hooks/useCioAutocomplete.ts
+++ b/src/hooks/useCioAutocomplete.ts
@@ -75,6 +75,7 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
     advancedParameters,
     defaultInput,
     getSearchResultsUrl,
+    onIsOpenChange,
   } = memoizedOptions;
 
   let { sections = defaultSections, zeroStateSections } = memoizedOptions;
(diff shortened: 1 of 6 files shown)


src/hooks/useCioAutocomplete.ts
```diff
@@ -125,7 +126,7 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
     highlightedIndex,
     getInputProps,
     getItemProps: getItemPropsDownShift,
-  } = useDownShift({ setQuery, items, onSubmit, cioClient, previousQuery });
+  } = useDownShift({ setQuery, items, onSubmit, cioClient, previousQuery, onIsOpenChange });
```
ZSnake, line 129: Could it be an issue that the `isOpen` value we're exposing _might be_ different from downshift's internal state? Since we return `isOpen: isOpen && items?.length > 0` there is a scenario where `onIsOpenChange` can trigger with `changes.isOpen` being true and our own `isOpen` as false.

ZSnake (approved):
Small comment, but looks good!

src/hooks/useCioAutocomplete.ts
mocca102: This is a great catch! Thank you for raising it

mocca102: `isOpen && items?.length > 0` was added to not show the dropdown if there are no items

I just got an ask from a customer to be able to show the dropdown even if there's no items.

I will create a separate story for that

PR #187, opened 2025-01-03
[CI-4057] Add no results message

stanlp1:


src/hooks/useCioAutocomplete.ts
```diff
@@ -141,6 +142,7 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
     request,
     featureToggles: features,
     isOpen: isOpen && items?.length > 0,
+    isFocused: isOpen || inputFocusState,
```
stanlp1, line 145: I believe isOpen is effectively the input focus state but I wasn't sure.

src/components/Autocomplete/AutocompleteResults/AutocompleteResults.tsx
```diff
@@ -39,23 +40,35 @@ const DefaultRenderResults: RenderResults = ({ sections }) =>
...
   let content;
+
   if (isOpen && hasResults) {
     content = (
       <>
         <CloseIcon onClick={() => closeMenu()} />
         {typeof children === 'function' ? children({ sections, getItemProps }) : children}
       </>
     );
+  } else if (query?.length && isFocused && displayNoResultsMessage) {
```
mocca102, line 69: This wrongly shows the "no results" message 

- When the results are still loading from the API while you're typing
- When you type something then click escape
- When you select a suggestion

mocca102 (changes requested):
There are some edge cases that need to be covered with when the "no results" is to be rendered

src/hooks/useCioAutocomplete.ts
mocca102: I believe yes

src/components/Autocomplete/AutocompleteResults/AutocompleteResults.tsx
mocca102: What if we check that the API response is empty to decide whether to show the message or not?

src/hooks/useCioAutocomplete.ts
mocca102: I believe you added `isFocused` becuase our `isOpen: isOpen && items?.length > 0`

I think we need to change that since the dropdown can be open now even though `items?.length === 0` so it's not accurate anymore

mocca102: If we did that we won't need to add `isFocused`

stanlp1: Yea that is why. If it just use isOpen it will solve the edge cases you mentioned in the other comment.

src/components/Autocomplete/AutocompleteResults/AutocompleteResults.tsx
stanlp1: > What if we check that the API response is empty to decide whether to show the message or not?

We allow "custom section" so we I don't think we can just check the api response by itself. I will see if there is another way to address the issue when the api is loading.

src/hooks/useCioAutocomplete.ts
stanlp1: But I am not sure if removing items?.length check will affect anything else, if a customer has behavior that relies on this logic 🤔

src/components/Autocomplete/AutocompleteResults/AutocompleteResults.tsx
mocca102: Can we do no results per section? 🤔 

So the message appears if there's no Search Suggestion side by side with product suggestions for example

src/hooks/useCioAutocomplete.ts
mocca102: It probably will. Lets discuss with the broader team

I think we can still make the change and mark it breaking maybe

src/components/Autocomplete/AutocompleteResults/AutocompleteResults.tsx
stanlp1: We probably can add add this configuration under each section and do it that way.

stanlp1: Made the changes to do it per section, I think that addresses all the issues.

mocca102 (changes requested):
@stanlp1  In the storybook example the message shows for the search suggestions sections but not for the Products suggestions. The products section just disappears
Screenshot 2025-01-15 at 1 39 10 PM 
Screenshot 2025-01-15 at 1 37 47 PM

mocca102 (approved):
LGTM!

PR #189, opened 2025-01-05
Undo memoizing all of props options and only memoize Objects/Arrays

mocca102:
**Problem Statement**

1. We [memoized all props](https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/158) to fix an issue with infinite rerenders when sections or zeroStateSections are passed directly to useCioAutocomplete as a literal array
2. This memoization introduced another issue where callback functions become stale

**Definition of Done:**

- Callback functions like `onSubmit` should have access to the most up-to-date values
- Passing object and array literals to `useCioAutocomplete` should not cause an infinite rerender

**How to reproduce:**

- add this to FullExampleTemplate in src/stories/Autocomplete/Component/index.TSX and then test on "Provide API Key" story
```
  const [x, setX] = useState('123');
  console.log(x); // This will have the most up to date value

  const onSubmit = () => {
    setX('456');
    console.log(x); // This will always log `123` even after x gets upated
  };
  
  return <CioAutocomplete {...args} onSubmit={onSubmit} />;
 

```

### DIFF
--- a/src/hooks/useNormalizedProps.ts
+++ b/src/hooks/useNormalizedProps.ts
@@ -1,0 +1,83 @@
+/* eslint-disable react-hooks/exhaustive-deps */
+import { useMemo } from 'react';
+import { UseCioAutocompleteOptions, UserDefinedSection } from '../types';
+import { isAutocompleteSection, isRecommendationsSection } from '../typeGuards';
+
+export const defaultSections: UserDefinedSection[] = [
+  {
+    indexSectionName: 'Search Suggestions',
+    type: 'autocomplete',
+  },
+  {
+    indexSectionName: 'Products',
+    type: 'autocomplete',
+  },
+];
+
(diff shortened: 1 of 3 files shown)


src/hooks/useNormalizedProps.ts
```diff
@@ -0,0 +1,83 @@
+/* eslint-disable react-hooks/exhaustive-deps */
```
ZSnake, line 1: I've seen this disabled in a bunch of places. Might be time to disable it permanently (or change it to warning instead) in the general eslint config.

ZSnake (approved):

src/hooks/useNormalizedProps.ts
mocca102: Your argument is getting stronger each time I add the comment to disable it 😄 I created a story for this "CI-4116"

PR #190, opened 2025-01-20
Bring back E2E tests to local env

mocca102:
**Problem statement:**

- E2E tests were removed by mistake
- Bring them back for local enviroment
- Some tests are failing but this will be handled in a different PR

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,5 +1,9 @@
+const isProduction = process.env.NODE_ENV === 'production';
+
 module.exports = {
-  stories: ['../src/**/Autocomplete/**/*.stories.mdx', '../src/**/Autocomplete/**/*.stories.@(js|jsx|ts|tsx)'],
+  stories: isProduction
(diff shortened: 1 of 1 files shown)


.storybook/main.js
```diff
@@ -1,5 +1,9 @@
+const isProduction = process.env.NODE_ENV === 'production';
+
 module.exports = {
-  stories: ['../src/**/Autocomplete/**/*.stories.mdx', '../src/**/Autocomplete/**/*.stories.@(js|jsx|ts|tsx)'],
+  stories: isProduction
```
ZSnake, line 4: Does this mean in production only autocomplete stories are tested?

ZSnake (approved):

.storybook/main.js
mocca102: Yes that is correct. The tests will be accessible only locally. It would be better if we can figure a way to enable it for prod as well it would be great but for now this is for enabling them locally

PR #193, opened 2025-01-30
Fallback to Merchandisers displayName set for recommendation pods

mocca102:
Before: 

- If displayName is not passed to the section configuration we used the pod ID as a fallback displayName

Screenshot 2025-01-30 at 5 43 50 PM

After:
- If displayName is not passed to the section configuration, we fall back to the displayName set by merchandisers and returned in the API response 
Screenshot 2025-01-30 at 5 45 10 PM

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -72,7 +72,8 @@ export const sectionsDescription = `- by default, typing a query will fetch data
 - the order of the objects in the \`sections\` array determines the order of the results
 - each autocomplete section object must have a \`indexSectionName\`
 - each recommendation section object must have a \`podId\`
-- each custom section object must have a \`displayName\`
+- recommendation section displayName if passed will override the displayName set by Merchandisers
+- each custom section object must have a \`displayName\`. 
 - each section object can specify a \`type\`
 - each section object can override the default \`numResults\` of 8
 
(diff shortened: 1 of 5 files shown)


src/constants.ts
```diff
@@ -72,7 +72,8 @@ export const sectionsDescription = `- by default, typing a query will fetch data
 - the order of the objects in the \`sections\` array determines the order of the results
 - each autocomplete section object must have a \`indexSectionName\`
 - each recommendation section object must have a \`podId\`
-- each custom section object must have a \`displayName\`
+- recommendation section displayName if passed will override the displayName set by Merchandisers
```
Mudaafi, line 75: wdyt of 

`passing displayName in the sectionConfiguration will override the displayName set by Merchandisers on the Customer Dashboard`

src/constants.ts
```diff
@@ -112,6 +113,7 @@ export const zeroStateDescription = `- when the text input field has no text, we
 - the order of the objects in the \`zeroStateSections\` array determines the order of the results
 - each autocomplete section object must have a \`indexSectionName\`
 - each recommendation section object must have a \`podId\`
+- recommendation section displayName if passed will override the displayName set by Merchandisers
```
Mudaafi, line 116: same as above

src/constants.ts
```diff
@@ -72,7 +72,8 @@ export const sectionsDescription = `- by default, typing a query will fetch data
 - the order of the objects in the \`sections\` array determines the order of the results
 - each autocomplete section object must have a \`indexSectionName\`
 - each recommendation section object must have a \`podId\`
-- each custom section object must have a \`displayName\`
+- recommendation section displayName if passed will override the displayName set by Merchandisers
+- each custom section object must have a \`displayName\`.
```
Mudaafi, line 76: extra fullstop (`.`)

src/constants.ts
mocca102: I approve

PR #196, opened 2025-03-04
[CI-4083] Enable forcing into testing cells for UI A/B tests

ZSnake:
# Changes
- Change prettier config match eslint config on trailing commas to avoid mismatch on formatting code
- Split up `utils.ts` function into several modules based on module responsibility
  -   `features`, `format`, `helpers`, `storage` and `tracking` as new util modules and moved storage related functions from `beaconUtils` to its own `storage` module so it can be used by other elements
- Add util functions in `features` module to read custom UI variant from `sessionStorage` and add logic to `getFeatures` function to prioritize variant from storage.
- Add integration tests to check with and without `sessionStorage` variant scenario

# How to test
1. In Storybook, navigate to an example with no Advanced Parameters set to alter the component's features (for example, any of the top-level examples in the Component's story `?path=/docs/autocomplete-component--docs`). Perform a search to make sure the default component is being shown.
2. Through the chrome devtools, set a session storage `key:value` pair with the key `_constructorio_custom_autosuggest_ui_` and the value one of the available UI variants
```
'custom_autosuggest_ui_result_count'
'custom_autosuggest_ui_image'
'custom_autosuggest_ui_image_result_count'
'custom_autosuggest_ui_disable_recommendations_in_zero_state'
```
3. Refresh the page and test the same example as before. Make sure the expected UI variant is shown based on the value set in storage.


mocca102 (commented):
@ZSnake  Note that `displaySearchSuggestionImages` and `displaySearchSuggestionResultCounts` passed to `advancedParametrs` will always override `featureDisplaySearchSuggestionImages` and `featureDisplaySearchSuggestionResultCounts` from `getFeatures`

This logic lives [here](https://github.com/Constructor-io/constructorio-ui-autocomplete/blob/0acf5542a6a3d398f3582cac1a0a5bf7dcfa409f/src/components/Autocomplete/SectionItem/SearchSuggestionItem.tsx#L15) 

So you will not be able to force yourself into any experience if they already pass those as `true` or `false`

If this is ok then the PR is good

spec/Components/CioAutocomplete/CioAutocomplete.test.tsx
```diff
@@ -126,4 +126,63 @@ describe('CioAutocomplete Client-Side Rendering', () => {
...
+
+      options.forEach((option) => {
+        const suggestionCount = option.querySelector('.cio-suggestion-count');
+        expect(suggestionCount).not.toBeInTheDocument();
+      });
+    });
+
+    it('Takes custom UI variant from session storage and applies features based on the variant', async () => {
+      window.sessionStorage.setItem(
+        '_constructorio_custom_autosuggest_ui_',
```
mocca102, line 153: To be consistent with the storage naming convention can we remove the last `_`

spec/Components/CioAutocomplete/CioAutocomplete.test.tsx
```diff
@@ -126,4 +126,63 @@ describe('CioAutocomplete Client-Side Rendering', () => {
...
+        'custom_autosuggest_ui_result_count'
+      );
+
+      const props = {
```
mocca102, line 157: Can we add a test case with `advancedParamerers` to confirm the manually entered values does not override the `advancedParamerers`?

mocca102 (commented):
This is looking good. Just left small comments

mocca102:
I like splitting the utils into different files. Thank you for doing that

mocca102:
@ZSnake Bumping this

ZSnake:
> Note that displaySearchSuggestionImages and displaySearchSuggestionResultCounts passed to advancedParametrs will always override featureDisplaySearchSuggestionImages and featureDisplaySearchSuggestionResultCounts from getFeatures

Yeah I saw this too. Didn't change it because I think this should be the correct behaviour. If a parameter is explicitly set, it shouldn't be overriden by A/B tests or custom UIs. If the customer will be experimenting with the UI, we should just ask them to not explicitly set these params

spec/Components/CioAutocomplete/CioAutocomplete.test.tsx
```diff
@@ -126,4 +126,103 @@ describe('CioAutocomplete Client-Side Rendering', () => {
 
     expect(console.error).not.toHaveBeenCalled();
   });
+  describe('Custom UI variant', () => {
+    afterEach(() => {
+      window.sessionStorage.removeItem('_constructorio_custom_autosuggest_ui_');
```
mocca102, line 131: Will need to update this as well to `_constructorio_custom_autosuggest_ui`

spec/Components/CioAutocomplete/CioAutocomplete.test.tsx
```diff
@@ -184,5 +184,45 @@ describe('CioAutocomplete Client-Side Rendering', () => {
         expect(suggestionCount).toBeInTheDocument();
       });
     });
+
+    it('Applies features sent in advancedParameters even if a custom UI variant is set that affects that feature', async () => {
```
mocca102, line 188: The test is failing 🤔 


Screenshot 2025-03-18 at 9 04 27 PM

spec/Components/CioAutocomplete/CioAutocomplete.test.tsx
```diff
@@ -184,5 +184,45 @@ describe('CioAutocomplete Client-Side Rendering', () => {
...
+
+      fireEvent.change(searchInput, { target: { value: 'pants' } });
+
+      const options = (await screen.findAllByRole('option')).filter(
+        (elem) => elem.getAttribute('data-cnstrc-item-section') === 'Search Suggestions'
+      );
+
+      options.forEach((option) => {
+        const suggestionCount = option.querySelector('.cio-suggestion-count');
+        expect(suggestionCount).toBeInTheDocument();
```
mocca102: Shouldn't this check if it's NOT in the document?

ZSnake: Yeah I was testing my test lmao. I think you somehow you commented/tested between the time I figured out I pushed the wrong thing

mocca102 (approved):
LGTM!

PR #197, opened 2025-03-04
Override default blur behavior

mocca102:
**Before** 

- The dropdown menu closes on Blur

**After**
- The dropdown menu closes on Blur AND click is not inside '.cio-autocomplete' element or its children

### DIFF
--- a/src/hooks/useCustomBlur.ts
+++ b/src/hooks/useCustomBlur.ts
@@ -1,13 +1,13 @@
 import { useCallback, useEffect } from 'react';
 
-const useCustomBlur = (isOpen: boolean, closeMenu: () => void) => {
+const useCustomBlur = (isOpen: boolean, closeMenu: () => void, autocompleteClassName: string) => {
   const handleDocumentClick = useCallback(
     (event: MouseEvent) => {
-      if (isOpen && !(event.target as HTMLElement)?.closest('.cio-autocomplete')) {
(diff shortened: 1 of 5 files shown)


src/hooks/useCustomBlur.ts
```diff
@@ -0,0 +1,23 @@
+import { useCallback, useEffect } from 'react';
+
+const useCustomBlur = (isOpen: boolean, closeMenu: () => void) => {
+  const handleDocumentClick = useCallback(
+    (event: MouseEvent) => {
+      event.preventDefault();
+      event.stopPropagation();
```
stanlp1, line 7: This would potentially break any click events that are attached to the window 🤔 

Not sure what could possibly be affected by this but wonder if there is a way where we don't have to use preventDefault/stopPropagation to prevent any unintended behavior changes.

mocca102: Great callout

src/hooks/useCustomBlur.ts
```diff
@@ -0,0 +1,21 @@
+import { useCallback, useEffect } from 'react';
+
+const useCustomBlur = (isOpen: boolean, closeMenu: () => void) => {
+  const handleDocumentClick = useCallback(
+    (event: MouseEvent) => {
+      if (isOpen && !(event.target as HTMLElement)?.closest('.cio-autocomplete')) {
```
stanlp1, line 6: I think we need to use autocompleteClassName here in case they override.

stanlp1 (approved):

PR #205, opened 2025-04-15
Fix cio js client events not firing

mocca102:
The bug is that currently, you can’t listen to any of the dispatched events from the ConstructorIO client. Basically, this does not work.

`window.addEventListener('cio.client.autocomplete.getAutocompleteResults.completed', () => {});
`

This [line](https://github.com/Constructor-io/constructorio-client-javascript/blob/c280d29b1da6e04f5891045b9ace73c46b53576b/src/utils/event-dispatcher.js#L21) is the reason. waitForBeacon is true by default. Because we don’t load the beacon on OS UI AC the eventDispatcher remains inactive (active = false)

```
// If `waitForBeacon` option is set, only enable event dispatching once event is received from beacon
if (this.waitForBeacon) {
      this.active = false;
}
```

To fix the bug we just need to pass this to the cio instance
```
const cioClient = new ConstructorIOClient({
      ...
      eventDispatcher: { waitForBeacon: false },
});
```

To test the changes add this anywhere 

```
window.addEventListener('cio.client.autocomplete.getAutocompleteResults.completed', (ev) => {
     console.log('event', event)
});
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -63,7 +63,7 @@
         "node": ">=14"
       },
       "peerDependencies": {
-        "@constructor-io/constructorio-client-javascript": "^2.64.0",
+        "@constructor-io/constructorio-client-javascript": "^2.67.6",
         "downshift": "^7.2.1",
         "react": ">=16.12.0",
         "react-dom": ">=16.12.0",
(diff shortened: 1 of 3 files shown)


stanlp1 (approved):
Good catch. LGTM

PR #206, opened 2025-04-15
Fix custom blur

mocca102:
🐛  **The bug**

- The custom blur hook breaks the click events if and only uf `autocompleteClassName` consists of multiple class names, ie

    - `cio-autocomplete custom-autocomplete-styles` 

🔁  **How to reproduce the bug?**

- Selects are broken on the full featured [example](https://constructor-io.github.io/constructorio-ui-autocomplete/?path=/story/autocomplete-component--full-featured-and-styled-example) since it has 
 
    - `autocompleteClassName = 'cio-autocomplete custom-autocomplete-styles';`

✅ **The fix**
- Handle multiple class names by splitting and joining them again


Mudaafi (approved):
lgtm. I don't like how we're relying on a prop called `autocompleteClassName` to handle blurring since it's not intuitive if they used the Hooks Approach but that's a separate discussion to have. This fix looks good to go.

PR #207, opened 2025-04-21
Migrate storybook to v8 and its dependencies

mocca102:
# `Upgrading Storybook from version 7.6.20 to version 8.6.12..`

## **How?**

- `npx storybook@latest upgrade`

### 1. 🆙 Updated Node and NPM versions to LTS

- [Node 18+ is now required](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#dropping-support-for-nodejs-16)

### 2. ➕  Added 

 - `@storybook/manager-api` and `storybook/theming`
 - Migration tool outputted `Storybook packages used in your project, but they are missing from your project dependencies:`

### 3. 🔁 Replaced

- `@storybook/jest` and `@storybook/testing-library` with `@storybook/test` instead
- The Packages are known to be incompatible since Storybook 8

### 4. ➖  Removed 

- `argTypesRegex` since it's deprecated

### 5. Webpack compiler setup

- ➕  Added `@storybook/addon-webpack5-compiler-babel`
- Installed babel presets
- Webpack5 builder is now compiler agnostic, meaning you can choose a compiler addon that best fits your project
- I choose Babel

### 6. 🆙  Upgraded storybook dependencies

- `@storybook/<something>`

### 7. ➕  Added Visual Tests addon (It was recommended by the migration prompts)

- `@chromatic-com/storybook`

### 8. 🔁  Removed 

- ➖ `docs.autodocs` setting in `.storybook/main.js` since it is deprecated
- ➕ Added it to `.storybook/preview.jsx` instead


.storybook/preview.jsx
```diff
@@ -1,11 +1,11 @@
+import React from 'react';
```
mocca102, line 1: It was not required before but it breaks now if you don't import react

package.json
```diff
@@ -11,13 +11,11 @@
     },
     ".": {
       "import": "./lib/mjs/index.js",
-      "require": "./lib/cjs/index.js",
-      "types": "./lib/types/index.d.ts"
+      "require": "./lib/cjs/index.js"
```
mocca102, line 14: The types import was ignored here. When I updated esbuild it threw the warning.

We already export all the types globally [here](https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/207/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L28)

mocca102:
Interaction tests are only failing when ran in the CLI because of

> After upgrading Storybook, the interaction tests are failing because they are blocked from firing Constructor events by our this.humanity.isBot() check

src/stories/tests/ComponentTests.stories.tsx
```diff
@@ -37,8 +42,14 @@ const defaultArgs: CioAutocompleteProps = {
       indexSectionName: 'Products',
     },
   ],
+  ...explicitActionsSpies,
 };
 
+// @ts-ignore
+// eslint-disable-next-line
+window.navigator.__defineGetter__('webdriver', () => false);
```
mocca102, line 50: Override `webdriver` so our tracker module doesn't treat it as a bot and block tracking events

.github/workflows/test-build.yml
```diff
@@ -21,6 +21,8 @@ jobs:
           node-version: '18.13.0'
       - name: Install dependencies
         run: npm ci
+      - name: Install Playwright Browsers
+        run: npx playwright install --with-deps # Downloads browser binaries + OS dependencies
```
mocca102, line 25: Playwright is required to download browser binaries for the interaction tests

package.json
```diff
@@ -89,21 +93,23 @@
...
     "eslint-plugin-react": "^7.31.11",
     "eslint-plugin-react-hooks": "^4.6.0",
-    "eslint-plugin-storybook": "^0.6.12",
+    "eslint-plugin-storybook": "^0.12.0",
     "husky": "^8.0.1",
     "jest": "^29.7.0",
     "jest-environment-jsdom": "^29.7.0",
     "jest-fail-on-console": "^3.3.0",
     "license-checker": "^25.0.1",
+    "playwright": "^1.52.0",
```
mocca102, line 102: Needed for the interaction tests

esezen:
Are you also getting this while doing `npm run dev`? Should we look into it?
Screenshot 2025-04-24 at 9 45 27 PM

mocca102:
> Are you also getting this while doing `npm run dev`? Should we look into it?
>...

Thank you for raising this @esezen

Yes, I do get it as well. This is happening because of the current Component Story Format (CSF). They need to be migrated to the updated CSF but this is having no real effect right now other than the warnings

This is an example of the changes needed to migrate to the new CSF format. Currently, we only have the `index.stories.tsx` files in the updated CSF format.
Screenshot 2025-04-24 at 11 04 15 PM

PR #212, opened 2025-05-02
[CI-4320] Create a Props.mdx for CioAutocomplete

Mudaafi:
This PR creates a Props table for the `CioAutocomplete` table in a platform agnostic way. This is done through type-annotations, and can be migrated off Storybook through the use of [react-docgen](https://github.com/reactjs/react-docgen?tab=readme-ov-file) and/or [react-props-md-table](https://github.com/TeamWertarbyte/react-props-md-table?tab=readme-ov-file).

A new folder is created to conform to the new, recommended [Storybook Project Hierarchy](https://constructor.slab.com/posts/os-ui-api-reference-guide-708f4fb4#hu0rw-storybook-file-hierarchy). The old docs are left untouched in case other engineers are in the process of migrating them and can be merged in a future PR.

### Possible Enhancements
Currently, the "default values" column is left unfilled since `CioAutocomplete` doesn't have default values for these props. We can either have the default prop values explicitly set, or add some css to hide that column entirely.

We might want to add an "Advanced Parameters" table as well.

### DIFF
--- a/src/stories/components/CioAutocomplete/CioAutocomplete.stories.tsx
+++ b/src/stories/components/CioAutocomplete/CioAutocomplete.stories.tsx
@@ -1,0 +1,30 @@
+import React from 'react';
+import type { Meta, StoryObj } from '@storybook/react';
+import { CioAutocomplete, CioAutocompleteProps } from '../../../index';
+import { componentDescription, apiKey } from '../../../constants';
+
+const meta: Meta<typeof CioAutocomplete> = {
+  title: 'Components/CioAutocomplete',
+  component: CioAutocomplete,
+  parameters: {
+    docs: {
+      description: {
+        component: componentDescription,
+      },
+    },
+  },
+};
+
+export default meta;
+type Story = StoryObj<typeof meta>;
+
+function StoryTemplate({ args }: { args: CioAutocompleteProps }) {
+  return <CioAutocomplete {...args} />;
+}
+
+export const PrimaryStory: Story = {
+  render: (args) => <StoryTemplate args={args} />,
+  args: {
+    apiKey,
+  },
+};
--- a/src/types.ts
+++ b/src/types.ts
@@ -15,8 +15,17 @@ import {
 export type { IAutocompleteParameters } from '@constructor-io/constructorio-client-javascript/lib/types';
 
 export type CioClientConfig = {
+  /**
+   * Your Constructor API key. Either `apiKey` or `cioJsClient` are required
+   */
   apiKey?: string;
+  /**
+   * Optional custom constructor instance. Either `apiKey` or `cioJsClient` are required
+   */
   cioJsClient?: ConstructorIOClient;
+  /**
+   * If you don't want to create an instance of the `ConstructorIOClient` but still want to customize some of the options, you can pass a `cioJsClientOptions` object. You can learn more about the possible values [Here](https://constructor-io.github.io/constructorio-client-javascript/ConstructorIO.html)
+   */
   cioJsClientOptions?: ConstructorClientOptions;
 };
 
@@ -35,17 +44,55 @@ export type AdvancedParameters = AdvancedParametersBase &
   Omit<IAutocompleteParameters, 'resultsPerSection'>;
 
 export type CioAutocompleteProps = CioClientConfig & {
+  /**
+   * Set to `false` to show suggestions only after a user clears their query,
+   * but not when they initially select the input
+   */
   openOnFocus?: boolean;
+  /**
+   * Function that takes in a Search Suggestion, and returns a url to redirect to
+   */
   getSearchResultsUrl?: (item: SearchSuggestion) => string;
+  /**
+   * Callback function that runs when the user submits a search.
+   * Usually used to trigger a redirect
+   */
   onSubmit: OnSubmit;
+  /**
+   * Callback function that runs when the user focuses on the input
+   */
   onFocus?: () => void;
+  /**
+   * Callback function that runs when the user modifies input
+   */
   onChange?: (input: string) => void;
+  /**
+   * Search input placeholder
+   */
   placeholder?: string;
+  /**
+   * Children to be rendered according to the RenderProps pattern
+   */
   children?: ReactNode;
+  /**
+   * Override the sections that will be rendered. Defaults to Products and Search Suggestion sections
+   */
   sections?: UserDefinedSection[];
+  /**
+   * Custom sections that should be rendered during the autocomplete zero state
+   */
   zeroStateSections?: UserDefinedSection[];
+  /**
+   * Override the parent container's class. Defaults to `cio-autocomplete`
+   */
   autocompleteClassName?: string;
+  /**
+   * See Advanced Parameters
+   */
   advancedParameters?: AdvancedParameters;
+  /**
+   * Search input default value
+   */
   defaultInput?: string;
 };
 
--- a/src/stories/components/CioAutocomplete/CioAutocompleteProps.mdx
+++ b/src/stories/components/CioAutocomplete/CioAutocompleteProps.mdx
@@ -1,0 +1,8 @@
+import { Meta, Markdown, ArgTypes } from '@storybook/blocks';
+import * as CioAutocompleteStories from './CioAutocomplete.stories';
+
+<Meta of={CioAutocompleteStories} />
+
+## Props
+
+<ArgTypes />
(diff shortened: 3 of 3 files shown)


src/stories/components/CioAutocomplete/CioAutocomplete.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { CioAutocomplete, CioAutocompleteProps } from '../../../index';
+import { componentDescription, apiKey } from '../../../constants';
+
+const meta: Meta<typeof CioAutocomplete> = {
+  title: 'Components/CioAutocomplete',
+  component: CioAutocomplete,
+  parameters: {
+    docs: {
+      description: {
+        component: componentDescription,
```
mocca102, line 12: I vote for removing/editing this description into a single line instead of bullet points

But won't hurt if we leave it as is.

src/types.ts
```diff
@@ -35,17 +44,55 @@ export type AdvancedParameters = AdvancedParametersBase &
...
   autocompleteClassName?: string;
+  /**
+   * See Advanced Parameters
+   */
   advancedParameters?: AdvancedParameters;
+  /**
+   * Search input default value
+   */
   defaultInput?: string;
 };
```
mocca102, line 98: Love that!

mocca102 (commented):
I would like docs + props to be a single page named 'docs'. I saw [Adobe](https://opensource.adobe.com/spectrum-web-components/storybook/?path=/story/accordion--default) and [Circuit UI](https://5f27fec87f827f00226fd51a-mkzlawwyms.chromatic.com/?path=/docs/navigation-tabs--docs&globals=colorScheme:dark) doing this and I like it more.

Let me know what you think

src/types.ts
```diff
@@ -35,17 +44,55 @@ export type AdvancedParameters = AdvancedParametersBase &
...
 
 export type CioAutocompleteProps = CioClientConfig & {
+  /**
+   * Set to `false` to show suggestions only after a user clears their query,
+   * but not when they initially select the input
+   */
   openOnFocus?: boolean;
+  /**
+   * Function that takes in a Search Suggestion, and returns a url to redirect to
+   */
```
mocca102, line 54: > If passed, the "Search Suggestions" rendered as HTML items will be wrapped by an HTML link tag with `href` attribute that redirects to the selected query search results page. (e.g., `<a href='www.mywebsite.com/search?query=[selected_search_suggestion]'>[Search Suggestion]</a>`

I think we can make this a bit clearer. I'm sure you can phrase this better than me but I was thinking something like the above

src/stories/components/CioAutocomplete/CioAutocomplete.stories.tsx
Mudaafi: Yeah I was going to remove it but since I'm not modifying the actual docs page, I figured someone else can remove it :p

src/types.ts
```diff
@@ -15,8 +15,17 @@ import {
...
+  /**
+   * Your Constructor API key. Either `apiKey` or `cioJsClient` are required
+   */
   apiKey?: string;
+  /**
+   * Optional custom constructor instance. Either `apiKey` or `cioJsClient` are required
+   */
   cioJsClient?: ConstructorIOClient;
+  /**
+   * If you don't want to create an instance of the `ConstructorIOClient` but still want to customize some of the options, you can pass a `cioJsClientOptions` object. You can learn more about the possible values [Here](https://constructor-io.github.io/constructorio-client-javascript/ConstructorIO.html)
```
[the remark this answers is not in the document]
Mudaafi: +1, I copied this from the previous description but I like your idea

src/types.ts
Mudaafi: How does this sound? 
```
Transforms a `SearchSuggestion` into the desired URL string to be used when rendering anchor tags
```

src/types.ts
```diff
@@ -35,17 +44,55 @@ export type AdvancedParameters = AdvancedParametersBase &
...
+   * Children to be rendered according to the RenderProps pattern
+   */
   children?: ReactNode;
+  /**
+   * Override the sections that will be rendered. Defaults to Products and Search Suggestion sections
+   */
   sections?: UserDefinedSection[];
+  /**
+   * Custom sections that should be rendered during the autocomplete zero state
+   */
```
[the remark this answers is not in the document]
Mudaafi: Hmm, I feel like Zero State Autocomplete is a pretty clear-cut term, at least in Constructor-land. I wouldn't want to constantly define what zero-state is every time I want to use that term, so how about I add "Zero-State Autocomplete" to our [Glossary Page](https://docs.constructor.com/docs/faq-glossary), and just hyperlink that term so users who aren't aware of what it means can get a quick refresher?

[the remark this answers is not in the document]
Mudaafi: As for "Custom Sections" being ambiguous in terms of who's powering them, maybe something like:
```
Configure either Constructor-powered sections or power your own custom sections to be rendered during in the [Zero State Autocomplete experience](link-tbd)
```

Mudaafi:
> I would like docs + props to be a single page named 'docs'. I saw [Adobe](https://opensource.adobe.com/spectrum-web-components/storybook/?path=/story/accordion--default) and [Circuit UI](https://5f27fec87f827f00226fd51a-mkzlawwyms.chromatic.com/?path=/docs/navigation-tabs--docs&globals=colorScheme:dark) doing this and I like it more.

@mocca102 I think my main concern is our docs will look much more cluttered than theirs since we don't hide our code examples behind stories (see Screenshot). It's something we've moved away from since the "show code" function is not the best DX. Let's keep this format for now, and when we finish the Story for the actual docs page, we can see if we want to merge it or not.

Screenshot 2025-05-06 at 12 34 43 AM

src/types.ts
mocca102: Looks good. Do you think we can add an example like this?

> <a href='www.mywebsite.com/search?query=[selected_search_suggestion]'>[Search Suggestion]</a>

src/types.ts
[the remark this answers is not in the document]
mocca102: > so how about I add "Zero-State Autocomplete" to our [Glossary Page](https://docs.constructor.com/docs/faq-glossary), and just hyperlink that term so users who aren't aware of what it means can get a quick refresher?

Even better. I like that suggestion more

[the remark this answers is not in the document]
mocca102: > Configure either Constructor-powered sections or power your own custom sections to be rendered for [Autocomplete Zero-State](link-tbd)

Why we're only mentioning custom sections here though? We can add custom sections either for zero or non zero state

src/types.ts
```diff
@@ -120,7 +121,7 @@ export type AutocompleteSearchSubmit = {
  * @example if (isAutocompleteSelectSubmit(event)) { ... } // `query` is available
  */
 export const isAutocompleteSelectSubmit = (
-  event: AutocompleteSubmitEvent
+  event: AutocompleteSubmitEvent,
```
mocca102, line 124: the linter is failing here

mocca102 (approved):
LGTM! Feel free to merge after fixing the lint issues

I left some more small comments, if you can have a look at them it would be great if not they're not so important

src/types.ts
Mudaafi: How about something like this?

```
<a href=getSearchResultsUrl([selected_search_suggestion])>[Search Suggestion]</a>
```

Afraid they might get confused at what `getSearchResultsUrl` is supposed to return instead

src/types.ts
Mudaafi: Good catch, looks like the right configuration was `"trailingComma": "es5"` instead of "all"

src/types.ts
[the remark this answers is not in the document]
Mudaafi: updated the type definition for `sections` to mention custom sections

src/types.ts
mocca102: Yes this is better

mocca102:
Spell check is failing because of some node version incompatibility 🤔

Mudaafi:
> Spell check is failing because of some node version incompatibility 🤔

Yeah I think the GHA we have wasn't pinned to a major version? We used to use spellcheck v8 but now we're pulling v9 which requires node v20>. I think it should be safe to merge, and then we can pull up a separate PR to fix it.

mocca102:
Is [storybook-new-docs-beta](https://github.com/Constructor-io/constructorio-ui-autocomplete/tree/storybook-new-docs-beta) the correct branch to branch off? Or `docs-revamp` the one we're using for the readme updates? @Mudaafi

PR #219, opened 2025-05-03
[CI-4292] Automate Table of Contents Generation using Doctoc

Mudaafi:


.husky/pre-push
```diff
@@ -0,0 +1,19 @@
...
+  DIFF_CONTENT=$(git diff HEAD @{push} README.md | sed 1,4d)
+
+  # Look for added lines (+) outside the TOC section
+  DIFF_CONTENT_EXCLUDING_TOC=$(echo "$DIFF_CONTENT" | sed '/<!-- START doctoc/,/<!-- END doctoc/d' | grep '^\+' || true)
+
+  if [ -n "$DIFF_CONTENT_EXCLUDING_TOC" ]; then
+    echo "Changes detected outside TOC section. Running doctoc..."
+    npm run doctoc-readme
+    git add README.md
+    git commit --amend --no-edit
```
Mudaafi, line 17: This will automatically amend the commit that is about to be pushed to remote. It's seamless, so as a contributor you might not even realize your final commit has been changed before it was pushed. That said, it lowkey scares me so we could make it a manual process instead.

Mudaafi: Update: @esezen suggested maybe adding a new commit instead of amending the existing one, but it turns out that doesn't play well cc: https://stackoverflow.com/questions/21334493/git-commit-in-pre-push-hook

Mudaafi:
@esezen Hmm it works for me. Let's hop on a sync

Mudaafi:
@enes @mocca102 Following our discussion, since amending the commits in-flight when a `git push` is run leads to flaky behavior, I've modified the script to be a `post_commit` hook instead. Logic is similar, but now it adds a new, separate commit if the README.md changes result in a TOC update.

esezen (approved):
LGTM!

PR #224, opened 2025-06-06
[CI-4464] Add item ids to select tracking

stanlp1:


src/hooks/useDownShift.ts
```diff
@@ -57,10 +57,13 @@ const useDownShift: UseDownShift = ({
               // (ie: Search Suggestions, Products, Custom Cio sections, etc)
               // This does not apply to custom user defined sections that aren't part of Constructor index
             } else if (selectedItem.result_id) {
-              trackAutocompleteSelect(cioClient, selectedItem.value, {
+              const selectData: any = {
```
esezen, line 60: Can we either properly type this or inline it?

stanlp1: Don't remember why I added any. Removed it.

PR #226, opened 2025-07-18
[CI-4607] Change CODEOWNERS

Alexey-Pavlov:


Mudaafi (commented):
@esezen should we retain the old shepherds or do we throw a cdx tag here instead?

esezen:
> @esezen should we retain the old shepherds or do we throw a cdx tag here instead?

@Mudaafi Let's do old shepherds for now but we will revisit all OS repo shepherds soon

PR #231, opened 2025-09-24
[NOCI] Bugfix: Rendering More than Expected Number of Items

evanyan13:
### The Issue:
* When variation_slicing is in place, there will be multiple items in the response with the same `item_id`.
* This causes React to have trouble differentiating components, leading to unexpected rendering behavior.

### Key Changes
* Update key assignment to `SectionItem` to use `item?.data?.variation_id` as the primary key as `item.id` as the fallback
* This ensures uniqueness of key, whether variation_slicing is in place or not
* Removed `key` as an explicit prop in `SectionItemProps` and `SectionItemsListProps`
  * The `key` field is a special prop that is used by React internally and it is not being passed down to the component as part of the props object. Such, there is no need to define it in the component props explicitly.

### DIFF
--- a/src/components/Autocomplete/SectionItem/SectionItem.tsx
+++ b/src/components/Autocomplete/SectionItem/SectionItem.tsx
@@ -9,7 +9,6 @@ import SearchSuggestionItem from './SearchSuggestionItem';
 export interface SectionItemProps {
   item: Item;
   children?: ReactNode;
-  key?: string;
   displaySearchTermHighlights?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


src/components/Autocomplete/SectionItem/SectionItem.tsx
```diff
@@ -9,7 +9,6 @@ import SearchSuggestionItem from './SearchSuggestionItem';
 export interface SectionItemProps {
   item: Item;
   children?: ReactNode;
-  key?: string;
```
evanyan13, line 12: * Suggesting to remove `key` as an explicit prop in `SectionItemProps` and `SectionItemsListProps` as
  * The `key` field is a special prop that is used by React internally and it is not being passed down to the component as part of the props object. Such, there is no need to define it in the component props explicitly.

PR #268, opened 2026-03-18
[CDX-404] Update packages and jsdom version

esezen:
Ran `npm audit fix` to get rid of security vulnerabilities. I had to upgrade jsdom to version 30 which is a breaking change with major differences.

- Upgrade `jest-environment-jsdom` from `^29.7.0` to `^30.3.0` (pulls in `jsdom@26`)
- Fix all 18 test files that relied on `Object.defineProperty(window, 'location', ...)` which is no longer allowed in jsdom 26
- Add a custom jsdom environment to patch jsdom's navigation behavior

Why do we need a custom jsdom environment?

jsdom 26 does not support `window.location.href = '...'` — it throws `"Not implemented: navigation (except hash changes)"`. This is a long-standing jsdom limitation, but in previous versions our tests worked around it by replacing `window.location` entirely with a plain `URL` object (where `.href` is just a writable property, not a navigation trigger).

With jsdom 26 locking down `window.location`, we can no longer swap it out. Our source code (`urlHelpers.ts`, `shopifyDefaults.ts`) intentionally assigns to `window.location.href` to navigate, and multiple tests exercise that code path. Without the patch, those tests would fail with the "Not implemented: navigation" error.

Let me know if there is an easier approach to this 🤔

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -3,7 +3,7 @@ module.exports = {
   projects: [
     {
       displayName: 'client',
-      testEnvironment: 'jsdom',
+      testEnvironment: './spec/jest-env-jsdom.js',
       testMatch: ['**/**/*.test.(js|jsx|ts|tsx)', '!**/**/*.server.test.(js|jsx|ts|tsx)'],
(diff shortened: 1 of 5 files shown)


TarekAlQaddy (approved):
I guess we have to keep in mind the jsdom internal patching just in case it ever changes in future versions

PR #279, opened 2026-04-22
[CDX-443] Fix rex bug with pod ID conflicts

esezen:
Two recommendation pods with the same `podId` but different `indexSectionName` (e.g. `bestsellers` targeting both `Products` and `Search Suggestions`) were collapsing into a single result set because the internal cache was keyed by `podId` alone.

`useFetchRecommendationPod` wrote both `recommendationsResults` and `podsData` as `{ [podId]: ... }`. The second API response silently overwrote the first, and every downstream reader (helpers, `useActiveSections`, `useCioAutocomplete`, React keys, view tracking) inherited the collision.

- New `getRecommendationPodKey(podId, indexSectionName)` helper producing `${podId}::${indexSectionName ?? 'Products'}`, threaded through every write and read site.
- Fixed inverted guard in `useSections` (`if (!podsData)` → `if (podsData)`) so the local state actually hydrates.
- `trackRecommendationView` now disambiguates on `podId` + `indexSectionName` so view events attribute to the correct section.
- `data-cnstrc-section` on recommendation containers falls back to `indexSectionName` when the pod returns zero results.

## Breaking change
`podsData` returned from `useCioAutocomplete` is now keyed by the composite key instead of bare `podId`. External consumers reading this map by bare `podId` will need to update the lookup.

Public types, SDK request shape, and tracking payload are unchanged.

### DIFF
--- a/src/hooks/useCioAutocomplete.ts
+++ b/src/hooks/useCioAutocomplete.ts
@@ -314,8 +314,7 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
 
       // Add data attributes for recommendations using helper
       if (isRecommendationsSection(section)) {
-        const podData =
-          podsData?.[getRecommendationPodKey(section.podId, section.indexSectionName)];
+        const podData = podsData?.[getRecommendationPodKey(section)];
         const seedItems = normalizeSeedItems(section.itemIds);
 
         const recommendationAttributes = getRecommendationsSectionCnstrcDataAttributes(
--- a/src/hooks/useCioAutocomplete.ts
+++ b/src/hooks/useCioAutocomplete.ts
@@ -302,7 +302,9 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
         className: `cio-section cio-section-${sectionListingType} ${getDeprecatedClassNames()}`,
         ref: section.ref,
         role: 'none',
-        [cnstrcDataAttrs.common.section]: section.data[0]?.section,
+        [cnstrcDataAttrs.common.section]:
+          section.data[0]?.section ||
+          (isRecommendationsSection(section) ? section.indexSectionName : undefined),
       };
 
       if (isCustomSection(section)) {
--- a/src/hooks/useCioAutocomplete.ts
+++ b/src/hooks/useCioAutocomplete.ts
@@ -304,7 +304,7 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
         role: 'none',
         [cnstrcDataAttrs.common.section]:
           section.data[0]?.section ||
-          (isRecommendationsSection(section) ? section.indexSectionName : undefined),
+          (isCustomSection(section) ? undefined : section.indexSectionName),
       };
 
       if (isCustomSection(section)) {
--- a/src/hooks/useCioAutocomplete.ts
+++ b/src/hooks/useCioAutocomplete.ts
@@ -14,7 +14,11 @@ import {
   OnSubmit,
 } from '../types';
 import usePrevious from './usePrevious';
-import { getItemPosition, getItemsForActiveSections } from '../utils/helpers';
+import {
+  getItemPosition,
+  getItemsForActiveSections,
+  getRecommendationPodKey,
+} from '../utils/helpers';
 import { toKebabCase } from '../utils/format';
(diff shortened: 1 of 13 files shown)


esezen (commented):
Self review

src/hooks/useCioAutocomplete.ts
```diff
@@ -298,7 +302,9 @@ const useCioAutocomplete = (options: UseCioAutocompleteOptions) => {
         className: `cio-section cio-section-${sectionListingType} ${getDeprecatedClassNames()}`,
         ref: section.ref,
         role: 'none',
-        [cnstrcDataAttrs.common.section]: section.data[0]?.section,
+        [cnstrcDataAttrs.common.section]:
+          section.data[0]?.section ||
+          (isRecommendationsSection(section) ? section.indexSectionName : undefined),
```
Mudaafi: good for zero-results, but ideally the `Section` object should always have an `indexSectionName` if it's derived from a Constructor response. A little refactor would be required to guarantee this for all paths, so we can leave it for a future PR.

src/utils/helpers.ts
```diff
@@ -82,6 +83,9 @@ export const getCioClient = (apiKey?: string, cioClientOptions?: ConstructorClie
   return null;
 };
 
+export const getRecommendationPodKey = (podId: string, indexSectionName?: string) =>
+  `${podId}::${indexSectionName ?? DEFAULT_RECOMMENDATION_INDEX_SECTION}`;
```
Mudaafi, line 98: hmm we could potentially run into the same issue if pods are instantiated with filters or different terms/itemIds. For example, we could still have the pods:
```
1. { podId: bestsellers, indexSectionName: Products }
2. { podId: bestsellers, indexSectionName: Products, filters: [...] }
```
In this scenario, the would both share the same key.

Mudaafi: Adding other params might get plenty lengthy though 🤔 Maybe we can take in the entire section and pick the parameters and hash it hmm

src/utils/tracking.ts
```diff
@@ -22,11 +22,13 @@ export const trackRecommendationView = (
...
   if (target.dataset.cnstrcRecommendationsPodId) {
-    // Pull recommendations from activeSectionsWithData by podId surfaced on target
+    // Pull recommendations from activeSectionsWithData by podId + section, since a
+    // single podId may be reused across multiple sections (e.g. Products vs. Search Suggestions)
     const recommendationSection = activeSectionsWithData.find(
       (section) =>
         isRecommendationsSection(section) &&
-        section.podId === target.dataset.cnstrcRecommendationsPodId
+        section.podId === target.dataset.cnstrcRecommendationsPodId &&
+        (!target.dataset.cnstrcSection || section.indexSectionName === target.dataset.cnstrcSection)
```
Mudaafi: Let's just use the podKey as the differentiator here

Mudaafi (commented):
one comment for discussion, otherwise it looks good to merge

src/utils/helpers.ts
Mudaafi: We discussed, and we probably don't need that granularity now so we can keep the logic as is (`product-section`).

As a side note, how do you feel about letting `getRecommendationPodKey` take the whole `RecommendationSectionConfiguration` as a prop instead?

src/hooks/useCioAutocomplete.ts
esezen: https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/279/changes/cf51140ec147f02cea23d29b7c61c44d8057b1c3

src/utils/helpers.ts
esezen: Like this? https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/279/changes/2b8df2706f04aa12f09e01887174d7f78ec86d2b

src/utils/tracking.ts
esezen: https://github.com/Constructor-io/constructorio-ui-autocomplete/pull/279/changes/2b252c2a324bebb983280691a9b8a5d4d4646b2c

src/utils/helpers.ts
Mudaafi: yep!

